PR #22, opened 2021-12-27
Support shuffle for data type smaller than 32-bit (e.g., fp16)

JamesTheZ:


linearhit (approved):
The CR has been review internally.

PR #25, opened 2021-12-28
[Chore] change submodules to ssh access

fortianyou:
SSH access is perfered because:

- Using the key is more secure than using a password
- No repetitive authentication is required as with HTTPS


Yancey1989 (approved):
LGTM，it's better to create issue to explain why doting this task and then create a pull request to fix it :)

linearhit (approved):
LGTM

PR #57, opened 2022-01-05
[disc-cpu] [bugfix] fix a bug of cpu multi-thread task assigner

wyzero:
For a nested parallel for loop:
```
scf.parallel((iv1, iv2, ... ivn) in range(L1, L2, ... LN)) {
  ...
}
```
The task assigner partitions the loop as even as possible across M threads.

Original implementation fails to tackle the situations like following example.

Suppose we have parallel domain: (5, 1) and 3 threads.

Original partition plan:

thread 0:
  - task 0: (4:5, 0:1),

thread 1:
  - task 0: (3:4, 0:1)
  - task 1:  (4:5, 0:1)  # this is illegal

thread 2:
  - task 0: (2:3, 0:1)
  - task 1:  (4:5, 1:2) # this is illegal

This PR solve the above problem.


Yancey1989 (commented):
Do we have a unit test to ensure the plan correct?

wyzero:
> Do we have a unit test to ensure the plan correct?

Add some tests. please take a look again. Thanks.

Yancey1989 (approved):
LGTM

PR #65, opened 2022-01-17
[Torch] Add DISC CPU CI

fortianyou:


pytorch_blade/tests/mlir/test_disc_nn_ops.py
```diff
@@ -64,6 +65,7 @@ def log_softmax_func(x):
             log_softmax_func, x=torch.randint(-3, 3, [2, 3, 10, 4], dtype=torch.int32)
         )
 
+    @unittest.skipIf(not cuda_available, "Please fix incorrect results")
```
wyzero, line 68: Could you please provide more info about this failure？ Is it totally wrong or just less precision? Under the hood, the compiler itself uses polynomial approximation for the `mhlo.exp` , which may lead to some precision loss.

.github/workflows/cpu_reusable.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+    # Runs a single command using the runners shell
+    - name: Checkout
+      uses: actions/checkout@v2.4.0
+    - name: Build Dev Docker
+      run: |
+        set -e
+        git submodule sync
+        git submodule update --depth=1 --init --recursive
+        docker build -t disc-dev-cpu --build-arg BASEIMAGE=nvidia/cuda:10.0-cudnn7-devel-ubuntu18.04 -f docker/dev/Dockerfile .
```
qiuxiafei, line 27: CPU test with cuda image? is this still the issue introduced by TF ?

qiuxiafei, line 27: image tag has a fixed name, any problem when simultaneously execution of multiple tasks?

scripts/ci/test_pytorch_blade.sh
```diff
@@ -29,8 +29,10 @@ export TF_REMOTE_CACHE=${TF_REMOTE_CACHE}
 # note(yancey.yx): using virtualenv to avoid permission issue on workflow actions CI,
 python -m virtualenv venv && source venv/bin/activate
 
+export TORCH_BLADE_CI_BUILD_TORCH_VERSION=${TORCH_BLADE_CI_BUILD_TORCH_VERSION:-1.7.1+cu110}
 (cd pytorch_blade \
-  && python -m pip install -q -r requirements-dev-1.7.1+cu110.txt -f https://pai-blade.oss-cn-zhangjiakou.aliyuncs.com/pytorch/wheels/repo.html  \
+  && python -m pip install -q -r requirements-dev-${TORCH_BLADE_CI_BUILD_TORCH_VERSION}.txt \
+       -f https://pai-blade.oss-cn-zhangjiakou.aliyuncs.com/pytorch/wheels/repo.html \
```
qiuxiafei, line 35: How about depending on official pypi index?

.github/workflows/cpu_reusable.yml
fortianyou: Nope, the workflow configuration was committed previously. I refactor it into a reusable workflow in order to share it between TensorFlow and PyTorch.

fortianyou: I think it won't be a problem, cause all uses of the image share the same Dockerfile.

scripts/ci/test_pytorch_blade.sh
fortianyou: The index-url might be chosen because it's stable? Anyway, this PR only do refactoring. It should be updated in another PR if the official index is preferred.

.github/workflows/cpu_reusable.yml
qiuxiafei: what if some changes were made into Dockerfile?

pytorch_blade/tests/mlir/test_disc_nn_ops.py
fortianyou:
```text
======================================================================
FAIL: test_softmax_func (tests.mlir.test_disc_nn_ops.TestDiscNNOps)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tianyou.gty/develop/opensource/BladeDISC/pytorch_blade/tests/mlir/test_disc_nn_ops.py", line 79, in test_softmax_func
    self._test_nn_ops(softmax)
  File "/home/tianyou.gty/develop/opensource/BladeDISC/pytorch_blade/tests/mlir/test_disc_nn_ops.py", line 23, in _test_nn_ops
    self._test_cvt_to_disc(nn_ops_func, test_data)
  File "/home/tianyou.gty/develop/opensource/BladeDISC/pytorch_blade/tests/mlir/testing_utils.py", line 54, in _test_cvt_to_disc
    self.assertEqual(output, result, rtol=rtol, atol=atol)
  File "/home/tianyou.gty/develop/opensource/BladeDISC/pytorch_blade/torch_blade/testing/common_utils.py", line 96, in assertEqual
    return super().assertEqual(x, y, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/torch/testing/_internal/common_utils.py", line 1165, in assertEqual
    super().assertTrue(result, msg=self._get_assert_msg(msg, debug_msg=debug_msg))
AssertionError: False is not true : Tensors failed to compare as equal!With rtol=1e-06 and atol=0.001, found 301056 element(s) (out of 301056) whose difference(s) exceeded the margin of error (including 301056 nan comparisons). The greatest difference was nan (0.004464285913854837 vs. nan), which occurred at index (0, 0, 0, 0).
```

qiuxiafei (approved):
LGTM

qiuxiafei (approved):

PR #75, opened 2022-01-24
Fix multi-definition error when building mlir_ral with cpu_only set.

JamesTheZ:


JamesTheZ:
It looks like the bug comes from environment configuration. It means only some environment configuration can trigger this problem. We will hold this change and do not merge currently.

JamesTheZ:
After an offline discussion, we decide to fix this problem through environment setting side modification, rather than the CPP code here. Closed.

PR #88, opened 2022-02-08
[torch converter][bugfix] Not convert transposed conv op

wyzero:
Original implementation puts a conv op into the supported list of DISC
no matter what the input values/attributes of the conv op are. However,
we actually do not support transposed conv op a.t.m. This CL fix the
problem.

### DIFF
--- a/pytorch_blade/src/compiler/mlir/converters/impl/convolution.cpp
+++ b/pytorch_blade/src/compiler/mlir/converters/impl/convolution.cpp
@@ -50,12 +50,16 @@ bool ConvertAtenConvolution(
   if (!CheckConstAttribute(jit_dilation, op_name, "dilation")) {
     return false;
   }
-  // other inputs is ignored currently
-  /*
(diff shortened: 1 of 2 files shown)


pytorch_blade/src/compiler/mlir/converters/impl/convolution.cpp
```diff
@@ -50,12 +50,16 @@ bool ConvertAtenConvolution(
...
     return false;
   }
-  // other inputs is ignored currently
-  /*
   auto jit_trans = node.input(6);
   if (!CheckConstAttribute(jit_trans, op_name, "transposed")) {
     return false;
   }
+  // TODO(disc): support transposed convolution.
+  if (CastJitConstToBool(*jit_trans)) {
```
fortianyou, line 58: Nit: Add an UnitTest

fortianyou (approved):
LGTM!

PR #89, opened 2022-02-08
[disc-cpu] support conv1d and conv3d

wyzero:
1, promote conv1d to conv2d by default.  This is the default behavior of TensorFlow. I also find that the promoted version is faster (+15%) than the original 1d version if the backend is oneDNN in some micro-benchmarks. 

2, add a 3d conv registration.

3, enable the conv1d & conv3d test of torch converter on cpu.


linearhit (approved):
LGTM

linearhit (approved):

linearhit (approved):

Issue #92, opened 2022-02-14
git am error when build with cxx11 abi


Orion34-lanbo:
In scripts/python/tao_build.py, when build with **cxx11 abi**, a git patch is applied. However this patch is **only used for gcc 7.3.0 when build with cxx11 abi**. Currently, we build cxx11 abi version with gcc 7.5.0 in ubuntu 18.04. We do not need it anymore,thus remove it.

Orion34-lanbo:
fixed

PR #93, opened 2022-02-14
[CI] remove useless git am action when build with cxx11 abi (#92)

Orion34C:


scripts/python/tao_build.py
```diff
@@ -471,9 +471,6 @@ def bazel_build(target, flag=""):
             execute("sed -i \
                 's/values\[i\] = internal::InnerMostDimReducer<Self, Op>::reduce(\*this, firstIndex + i \* num_values_to_reduce,$/CoeffReturnType t = internal::InnerMostDimReducer<Self, Op>::reduce(\*this, firstIndex + i \* num_values_to_reduce, num_values_to_reduce, reducer)\;values\[i\] = t\;/g' \
                 'bazel-tf_community/external/eigen_archive/unsupported/Eigen/CXX11/src/Tensor/TensorReduction.h'")
-        if args.ral_cxx11_abi:
-            # A workaround for build compiler main with gcc 730 when enable cxx11 abi
-            execute("git am ../platform_alibaba/ci_build/patch/workaround-to-fix-cxx11-build.patch")
```
linearhit, line 476: Do we need to remove the patch file from the repo as well?

linearhit (changes requested):

scripts/python/tao_build.py
Orion34C: The patch is not in this repo, this is what I want to fix at first, then I found it is not needed for cxx11 abi build, so just need to remove this logic from tao_build.py.

Yancey1989 (approved):
LGTM

linearhit (approved):
LGTM

PR #97, opened 2022-02-15
[Torch] add torch2mhlo for torch.roll and torch.unbind

fortianyou:


pytorch_blade/src/compiler/mlir/converters/impl/shapes.cpp
```diff
@@ -287,6 +287,82 @@ bool ConvertAtenSize(MhloConversionContext& ctx, const torch::jit::Node& node) {
...
+  mlir_dim_t dim = CastJitConstToInt64(*jit_dim);
+
+  auto tensor_type = jit_self->type()->cast<torch::TensorType>();
+  TORCH_CHECK(tensor_type != nullptr);
+
+  // NB: we get number of outputs from the Tensor's shape.
+  auto tensor_sizes = *(tensor_type->sizes().concrete_sizes());
+  auto dim_size = tensor_sizes[dim];
+  auto self = ctx.GetMlirValue(jit_self);
+  auto builder = *ctx.builder;
```
qiuxiafei, line 354: nit: maybe we don't have to create a copy of the builder? `BuildStdConstForI64` and `BuildSelect` just receive references.

qiuxiafei (approved):

PR #114, opened 2022-02-18
[Bazel][Torch] bazel build for TorchBlade

fortianyou:


wyzero:
Please rebase first since the CI was broken before.

fortianyou:
> 

done

pytorch_blade/.bazelversion
```diff
@@ -0,0 +1 @@
+4.2.2
```
wyzero, line 1: 5.0.0 since the rebase commit (20220216) is merged.

tao_compiler/mlir/disc/BUILD
```diff
@@ -1761,6 +1767,7 @@ transitive_hdrs(
         ":disc_map_chlo_to_hlo_op",
         ":mhlo_disc",
         "//tensorflow/compiler/mlir/hlo:hlo",
+        "//tensorflow/compiler/mlir/hlo:lhlo",
```
wyzero, line 1770: Why should we expose lmhlo header files?

tao_compiler/.bazelrc.user
```diff
@@ -0,0 +1,12 @@
+build:ci_build --noshow_loading_progress --show_progress_rate_limit=600 --jobs=10
+
+build:disc --define framework_shared_object=false --experimental_multi_threaded_digest
+build:disc_cpu --config=disc --config=release_cpu_linux --cxxopt=-DTAO_CPU_ONLY
+build:disc_cuda --config=disc --config=cuda
+build:disc_dcu --config=disc --config=dcu
+
+build:disc_mkldnn --cxxopt=-DTAO_ENABLE_MKLDNN --define is_mkldnn=true
```
wyzero, line 8: Should we need to apply `--config=disc` here as well?

scripts/python/tao_build.py
```diff
@@ -412,51 +302,43 @@ def bazel_build(target, flag=""):
...
+    if args.enable_blaze_opt:
+        configs.append('--config=disc_blaze')
+
+    if args.enable_mkldnn:
+        configs.append('--config=disc_mkldnn')
+
+    if running_on_ci():
+        configs.append('--config=ci_build')
+
+    BAZEL_BUILD_CMD = "bazel build --experimental_multi_threaded_digest --define framework_shared_object=false"
```
wyzero, line 323: could this be replaced with `--config=disc`?  or just simply remove these options since above config will also apply these?

pytorch_blade/bazel/torch/torch.BUILD.tpl
```diff
@@ -0,0 +1,77 @@
+package(default_visibility = ["//visibility:public"])
+
+load(
+    "@local_config_cuda//cuda:build_defs.bzl",
+    "if_cuda_is_configured",
+)
+
+cc_library(
+    name = "libtorch",
```
linearhit, line 9: I'm wondering if it's a proper usage of cc_library. 
Can we use copts with -L/-I instead of adding a cc_library "libtorch" ?

tao_compiler/mlir/disc/BUILD
linearhit: +1

tao_compiler/.bazelrc.user
```diff
@@ -0,0 +1,12 @@
...
+build:disc_blaze --cxxopt=-DBLAZE_OPT
+
+build:cxx11abi_0 --cxxopt=-D_GLIBCXX_USE_CXX11_ABI=0
+build:cxx11abi_1 --cxxopt=-D_GLIBCXX_USE_CXX11_ABI=1
```
linearhit, line 12: nit: at least one of them should be default?

pytorch_blade/cmake_build.py
```diff
@@ -17,98 +17,17 @@
 import shutil
 import re
 from distutils.spawn import find_executable
-import release_version
+from torch_blade_build import TorchBladeBuild, get_fullpath_or_create, check_env_flag
```
linearhit, line 20: what's the relationship between cmake_build.py and bazel_build.py
Will they both work for a while?

pytorch_blade/.bazelversion
fortianyou: done

pytorch_blade/bazel/torch/torch.BUILD.tpl
fortianyou: I think making the library a build target is better than using a link option. Because we can describe the dependencies in the target.

pytorch_blade/cmake_build.py
fortianyou: Yes, we currently support CMake and Bazel both.

scripts/python/tao_build.py
fortianyou: I deleted the parameters here because they are redundant. But, I would like to make disc configurable but not hard coding here.

tao_compiler/.bazelrc.user
fortianyou: Good point.

tao_compiler/.bazelrc.user
fortianyou: Actually, I think it's better to configure noting by default. And give the default setting of c++ compiler pass through.

linearhit (approved):
LGTM

tao_compiler/mlir/disc/BUILD
fortianyou: Deleted

qiuxiafei (approved):
Generally LGTM. 
I see many huge bazel targes which compiles all `.cc` file within a directory, but it's always a good practice to make targets smaller. It can speed up re-build and also reduce the possibility to link unwanted code for future targets.

tao_compiler/.bazelrc.user
fortianyou: There must be a conflict if both are configured.

mhlo_builder/csrc/BUILD
```diff
@@ -0,0 +1,26 @@
...
+filegroup(
+    name = "mhlo_headers",
+    srcs = glob(["mlir/mhlo/builder/*.h"])
+)
+
+cc_library(
+    name = "mhlo_library",
+    srcs = [":mhlo_sources"],
+    hdrs = [":mhlo_headers"],
+    includes = ["."],
```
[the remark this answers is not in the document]
fortianyou: Without this the gcc didn't no where to find cpp sources' header files

pytorch_blade/src/compiler/mlir/converters/BUILD
```diff
@@ -0,0 +1,46 @@
...
+filegroup(
+    name = "torch_blade_mhlo_converter_hdrs",
+    srcs = glob(["*.h"]),
+)
+
+cc_library(
+    name = "torch_blade_mhlo_converter",
+    srcs = [":torch_blade_mhlo_converter_srcs"],
+    hdrs = [":torch_blade_mhlo_converter_hdrs"],
+    includes = ["../../../"],
```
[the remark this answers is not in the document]
fortianyou: Good suggestion! But there are python files in the root directory as well. `src` directory only includes c/c++ sources/headers.

PR #126, opened 2022-02-24
ExpandDim-DynamicBroadcastInDim simplifier enhancement.

JamesTheZ:
Enhance the conversion from "ExpandDim - DynamicBroadcastInDim" to DynamicBroadcastInDim.

### DIFF
--- a/tao_compiler/mlir/disc/transforms/disc_shape_simplifier.cc
+++ b/tao_compiler/mlir/disc/transforms/disc_shape_simplifier.cc
@@ -276,14 +276,26 @@ class DynamicBroadcastInDimOpSimplifier
       }
 
       for (auto dim : fromElementsOp->getOperands()) {
-        auto staticDim =
-            dyn_cast_or_null<arith::ConstantIndexOp>(dim.getDefiningOp());
-        if (staticDim) {
+        // Deal with static dim.
+        int64_t staticVal = -1;
+        if (arith::ConstantIntOp constant_op =
+                dyn_cast_or_null<arith::ConstantIntOp>(dim.getDefiningOp())) {
+          staticVal = constant_op.getValue().cast<IntegerAttr>().getInt();
+          assert(staticValue > 0);
+        } else if (arith::ConstantIndexOp constant_op =
+                       dyn_cast_or_null<arith::ConstantIndexOp>(
+                           dim.getDefiningOp())) {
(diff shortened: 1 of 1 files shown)


tao_compiler/mlir/disc/transforms/disc_shape_simplifier.cc
```diff
@@ -202,7 +202,12 @@ class DynamicReshapeOpShapeInference
   }
 };
 
-// Match following patterns:
+// Convert an expand-like reshape op followed by a dynamic-broadcast-in-dim op
+// to a single dynamic-broadcast-in-dim op. The dim size of expanded dims should
+// be one.
+//
+// An example is like following:
```
linearhit, line 209: From the code, we are looking for a DynamicReshapeOp or ReshapeOp to match the pattern. But i didn't see any reshape here in the example?

tao_compiler/mlir/disc/transforms/disc_shape_simplifier.cc
```diff
@@ -212,56 +217,130 @@ class DynamicReshapeOpShapeInference
...
+    } else {
+      auto fromElementsOp = dyn_cast_or_null<tensor::FromElementsOp>(
+          dynReshapeOp->getOperand(1).getDefiningOp());
+      if (!fromElementsOp) {
+        return failure();
+      }
+
+      for (auto dim : fromElementsOp->getOperands()) {
+        auto staticDim =
+            dyn_cast_or_null<arith::ConstantIndexOp>(dim.getDefiningOp());
```
wyzero, line 280: What about `arith::ConstantIntOp`?

tao_compiler/mlir/disc/transforms/disc_shape_simplifier.cc
JamesTheZ: This example only shows a dynamic-reshape based pattern. ReshapeOp based pattern is similar except that the batch-dim value is constant folded.

linearhit (approved):
LGTM

tao_compiler/mlir/disc/transforms/disc_shape_simplifier.cc
JamesTheZ: Thanks for the comment. Fixed.

PR #128, opened 2022-02-28
UT for corner case of PTXAS error before v11.1.

JamesTheZ:
Add a UT about PTXAS compilation error before version of 11.1. Incase we encounter the same error in a future PTXAS release.

### DIFF
--- a/tao_compiler/mlir/disc/tests/regression/corner_case.cc
+++ b/tao_compiler/mlir/disc/tests/regression/corner_case.cc
@@ -1,0 +1,35 @@
+// Copyright 2022 The BladeDISC Authors. All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


tao_compiler/mlir/disc/tests/regression/corner_case.cc
```diff
@@ -0,0 +1,35 @@
...
+#include "tensorflow/compiler/mlir/disc/tests/mlir_feature_test.h"
+#include "tensorflow/compiler/mlir/disc/tests/mlir_test.h"
+#include "tensorflow/core/platform/test.h"
+
+namespace mlir_test {
+
+const std::string c_ft_path =
+    "tensorflow/compiler/mlir/disc/tests/regression/data/";
+
+#if CUDA_VERSION >= 11100
```
wyzero, line 21: Do we have this variable in in cpu-only env?

JamesTheZ: Then this UT will not execute because an undefined macro is replaced with 0.

linearhit: Should we provide a proper ptxas for users with 11.0?

JamesTheZ: > 

It is not the responsibility of the compiler itself. Even we provide a docker, we cannot prevent users to use other environments. (I do not think there will be that many users use our provided dockers by default.) Actually, the XLA engine will print warnings to complain about ptxas-11.0 if logging is enabled.

linearhit (approved):
LGTM

PR #129, opened 2022-02-28
[docs] add CPU support in build_from_source.md

### LINKED ISSUE #104: Missing cpu docker image and cpu build process in build_from_source.md

wyzero:
fix #104

### DIFF
--- a/docs/build_from_source.md
+++ b/docs/build_from_source.md
@@ -8,8 +8,8 @@ to build and test.
 ## Prerequisite
 
 1. Git for checking out the source code.
-1. [Nvidia Docker](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html)
-to launch Docker container on Nvidia GPU host.
+1. (Optional)[Nvidia Docker](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html)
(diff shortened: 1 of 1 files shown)


docs/build_from_source.md
```diff
@@ -46,19 +52,39 @@ development Docker image.
 development Docker image.
 
     ``` bash
+    # For GPU backend
     docker run --gpus all --rm -it -v $PWD:/disc bladedisc/bladedisc:latest-devel-cuda11.0 bash
+
+    # For CPU backend
+    docker run --rm -it -v $PWD:/disc bladedisc/bladedisc:latest-devel-cuda11.0 bash
```
linearhit, line 59: The commands are the same as for TensorFlow Users
We may move "docker run --rm -it -v $PWD:/disc bladedisc/bladedisc:latest-devel-cuda11.0 bash" to line19.

Thus this passage is no longer needed to be duplicated.
"""
**Note that we use the same development docker images for both Nvidia GPU backend and X86 CPU backend. For X86 CPU backend, both the building phase and execution phase do not require CUDA available.**
"""

linearhit (changes requested):

docs/build_from_source.md
wyzero: done

linearhit (approved):
LGTM

PR #130, opened 2022-02-28
Speculation strategy with better TLP for row-reduce.

JamesTheZ:
This provides a better row-reduction schedule speculation to increase ILP. Specifically, it decides to use more thread blocks when row-size is small and col-size is large enough.

### DIFF
--- a/tao_compiler/mlir/disc/transforms/codegen_utils.h
+++ b/tao_compiler/mlir/disc/transforms/codegen_utils.h
@@ -102,7 +102,7 @@ int getReductionTileSizeOnCPU();
 
 // SM number and max-threads-per-SM of different NVIDIA GPU architectures.
 // key: {arch-major, arch-minor}, value: {SM-number, max-threads-per-SM}
-const std::map<std::pair<int, int>, std::pair<int, int>> archToGPUThreadNumber =
+const std::map<std::pair<int, int>, std::pair<int, int>> ArchToGPUThreadNumber =
     {
         {std::make_pair(7, 0), std::make_pair(80, 2048)},   // V100
         {std::make_pair(7, 5), std::make_pair(40, 1024)},   // T4
--- a/tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
+++ b/tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
@@ -1,17 +1,13 @@
-/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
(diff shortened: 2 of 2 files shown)


tao_compiler/mlir/disc/transforms/codegen_utils.h
```diff
@@ -102,7 +102,7 @@ int getReductionTileSizeOnCPU();
 
 // SM number and max-threads-per-SM of different NVIDIA GPU architectures.
 // key: {arch-major, arch-minor}, value: {SM-number, max-threads-per-SM}
-const std::map<std::pair<int, int>, std::pair<int, int>> archToGPUThreadNumber =
+const std::map<std::pair<int, int>, std::pair<int, int>> ArchToGPUThreadNumber =
     {
         {std::make_pair(7, 0), std::make_pair(80, 2048)},   // V100
         {std::make_pair(7, 5), std::make_pair(40, 1024)},   // T4
```
wyzero, line 108: What if two GPUs have same SM capacity while have different SM number?

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
```diff
@@ -262,28 +258,65 @@ struct DiscSpecializeFusionWithSpeculationPass
...
+    // flexible per-reduce-thread-number setting.
+    bool experimental_tlp_enhance = false;
+    tensorflow::ReadBoolFromEnvVar("DISC_EXPERIMENTAL_SPECULATION_TLP_ENHANCE",
+                                   false, &experimental_tlp_enhance);
+    auto thread_number_info =
+        ArchToGPUThreadNumber.find({cc_major_, cc_minor_});
+    if (experimental_tlp_enhance &&
+        (thread_number_info != ArchToGPUThreadNumber.end())) {
+      // Experimental schedule selection policy is as following:
+      //   1. use schedule 1 if
```
wyzero, line 285: Is 'one block per row' better when both row size and column size are small?

JamesTheZ: Neither of existing 2 schedules can deal with the case you mentioned. This is why I leave a TODO in the comments about re-design the schedule according to `adaptive thread mapping` in AStitch.

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
```diff
@@ -479,17 +515,21 @@ struct DiscSpecializeFusionWithSpeculationPass
...
     Speculator(
         &DiscSpecializeFusionWithSpeculationPass::DoColReductionSpeculation);
 
-    // Stage #3: speculation of vectorization/tiling.
+    // Stage #3: speculation of row reduce op.
+    // We do row-reduction speculation before vectorization/tiling because TLP
+    // is usually more important than ILP on GPU.
+    Speculator(
+        &DiscSpecializeFusionWithSpeculationPass::DoRowReductionSpeculation);
+
```
linearhit, line 531: just curious, why do we swap the position of DoColReductionSpeculation & DoColReductionSpeculation?

JamesTheZ: It actually does not matter. I used to try to merge the row-reduction and vec/tiling speculations into a holistic one. And I realized that the merging is not necessary. Thus leave row-reduce speculation just upon the one for vec/tiling. Moreover, the comments added for row-reduce is related to vec/tiling, and putting them together is slightly more readable.

wyzero (approved):
LGTM

PR #140, opened 2022-03-03
[Torch] fix integer tensor div operation in torch 1.6

qiuxiafei:


pytorch_blade/tests/mlir/test_disc_tensor.py
```diff
@@ -53,7 +53,10 @@ def test_item(tensor):
 
         @torch.jit.script
         def test_item_2(tensor):
-            x = tensor / torch.tensor(2)
+            # Integer division of tensors using div or / is not supported only in torch
+            # 1.6, but works in 1.7 and 1.8, while the error message says it is
+            # supported until 1.6. So use // instead.
+            x = tensor // torch.tensor(2)
             x = int(x)
```
chenbohua3, line 60: the result of  operation // is already int, so `x = int(x)` could be deleted.

qiuxiafei: The result type is still `torch.tensor`
```
>>> torch.__version__
'1.8.0'
>>> type(torch.tensor(3) // torch.tensor(2))
<class 'torch.Tensor'>
```

Orion34-lanbo (approved):
LGTM

PR #142, opened 2022-03-04
Save const values in metadata.proto in splat mode

linearhit:
This change is to fix the problem that huge splat constant may exceed the limit of protobuf.

### DIFF
--- a/tao_compiler/mlir/xla/ral/context/common_context_impl_cuda.cc
+++ b/tao_compiler/mlir/xla/ral/context/common_context_impl_cuda.cc
@@ -75,7 +75,9 @@ static inline buffer_t ral_base_cuda_const_cuda_internal(
     TAO_VLOG(2) << "unique_name: " << key;
     auto it = state->device_constants.find(key);
     if (it == state->device_constants.end()) {
-      buffer_shape_t dim_sizes = GetShapeFromConstUniqueName(ctx, unique_name);
+      int64_t width_in_bytes = 0;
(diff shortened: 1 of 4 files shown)


tao_compiler/mlir/xla/ral/context/common_context_impl_cuda.cc
```diff
@@ -84,9 +86,19 @@ static inline buffer_t ral_base_cuda_const_cuda_internal(
...
-      TAO_VLOG(2) << "fromHex start:";
       auto data = fromHex(hex_str);
       auto bytes = data.size();
+      int64_t num_elements = std::accumulate(dim_sizes.begin(), dim_sizes.end(),
+                                             1, std::multiplies<int64_t>());
+      if (bytes < num_elements * width_in_bytes) {
+        // isSplat
+        bytes = num_elements * width_in_bytes;
+        auto splat_data = data;
+        for (int64_t i = 0; i < num_elements - 1; ++i) {
```
wyzero, line 97: NIT: maybe we could use memset on GPU？ For large const，it may be much faster.

wyzero (approved):
LGTM

tao_compiler/mlir/xla/ral/context/common_context_impl_cuda.cc
linearhit: memset api is only able to handle 32bit set, which is not enough.

Issue #113, opened 2022-02-18
Support fusing "isSplat" constants


linearhit:
The problem is observed in swin-transformer, when pytorch is doing amp.
```
"lmhlo.constant"(%275) {value = dense<0.000000e+00> : tensor<64x784x768xf32>} : (memref<64x784x768xf32, "gpu">) -> ()

…

"lmhlo.fusion"() ( {
      …
      "lmhlo.multiply"(%1741, %275, %1742) {disc.device = "gpu"} : (memref<64x784x768xf32, "gpu">, memref<64x784x768xf32, "gpu">, memref<64x784x768xf32, "gpu">) -> ()
      …
      "lmhlo.terminator"() : () -> ()
 }) {disc.device = "gpu", disc.fusion.name = "main_kLoop_reshape__37_1_2" : () -> ()

"lmhlo.fusion"() ( {
      …
      "lmhlo.multiply"(%1888, %275, %1889) {disc.device = "gpu"} : (memref<64x784x768xf32, "gpu">, memref<64x784x768xf32, "gpu">, memref<64x784x768xf32, "gpu">) -> ()
      …
      "lmhlo.terminator"() : () -> ()
 }) {disc.device = "gpu", disc.fusion.name = "main_kLoop_reshape__37_1_3" : () -> ()
```
In general, the "splat" constants outside of a fusion kernel might cause severe performance issues. In swin-transformer, the performance degradation can be very severe. Please be aware that there might be multiple kernels consuming the splat constant.

Solution 1: mark "splat" constant as fusible in fusion pass; and add an additional fusion stage that allows to duplicate the producer according to some forms of rules, like the FusionMerger in XLA. 

Solution 2: add an additional FuseSplatConstPass after the regular fusion pass that specifically duplicate and fuse the Splat const into fusion kernels.

Both solutions need also to support the fusion codegen for "splat" constants. Solution 2 can be regarded as a shrink version of solution 1， which can not handle such cases: 
```
"lmhlo.constant"(%272) {value = dense<0.000000e+00> : tensor<64x784x768xf32>} : (memref<64x784x768xf32, "gpu">) -> ()
"lmhlo.add"(%272, %273, %275)  {disc.device = "gpu"} : (memref<64x784x768xf32, "gpu">, memref<64x784x768xf32, "gpu">, memref<64x784x768xf32, "gpu">) -> ()

…

"lmhlo.fusion"() ( {
      …
[...]

PR #145, opened 2022-03-07
Add FuseSplatConstPass to support fusion of splat const to avoid

linearhit:
Add FuseSplatConstPass to support fusion of splat const to avoid unnecessary memory access #113

### DIFF
--- a/tao_compiler/mlir/disc/tests/regression/data/kstitch_fusion_with_splat_const.mlir
+++ b/tao_compiler/mlir/disc/tests/regression/data/kstitch_fusion_with_splat_const.mlir
@@ -1,0 +1,12 @@
+module attributes {tf.versions = {bad_consumers = [], min_consumer = 0 : i32, producer = 0 : i32}} {
+  func @main(%arg0: tensor<?x?xf32>) -> (tensor<?xf32>, tensor<?xf32>) attributes {tf.entry_function = {inputs = "{{INPUTS}}", outputs = "{{OUTPUTS}}", input_placements="{{INPUT_PLACEMENTS}}", output_placements="{{OUTPUT_PLACEMENTS}}"}} {
+    %graph:2 = tf_executor.graph {
+      %1:2 = tf_executor.island wraps "tf.Const"() {value = dense<[1]> : tensor<1xi32>} : () -> tensor<1xi32>
+      %2:2 = tf_executor.island wraps "tf.Max"(%arg0, %1) : (tensor<?x?xf32>, tensor<1xi32>) -> tensor<?xf32>
+      %3:2 = tf_executor.island wraps "tf.Const"() {value = dense<1.234> : tensor<f32>} : () -> tensor<f32>
+      %4:2 = tf_executor.island wraps "tf.Add"(%2, %3) : (tensor<?xf32>, tensor<f32>) -> tensor<?xf32>
+      tf_executor.fetch %2, %4 : tensor<?xf32>, tensor<?xf32>
+    }
+    return %graph#0, %graph#1 : tensor<?xf32>, tensor<?xf32>
+  }
+}
--- a/tao_compiler/mlir/disc/tests/regression/kstitch_fusion.cc
+++ b/tao_compiler/mlir/disc/tests/regression/kstitch_fusion.cc
@@ -133,4 +133,19 @@ TEST(KStitchFusionGPUTest, KStitch3DF64) {
   unsetenv("DISC_EXPECTED_KERNELS_IN_UT");
 }
 
+// With splat const in the kStitch fusion
+TEST(KStitchFusionGPUTest, KStitchSimpleWithSplatConstF32) {
+  setenv("DISC_ENABLE_STITCH", "true", 1);
+  setenv("DISC_EXPECTED_KERNELS_IN_UT", "1", 1);
+  EXPECT_TRUE(feature_test_main(
+      /*mlir_file_path*/ c_ft_path + "kstitch_fusion_with_splat_const.mlir",
+      /*backend_types*/ {BackendType::kCuda},
+      /*num_inputs*/ 1,
+      /*num_outputs*/ 2,
+      /*input_descriptors*/ {"11000x123xf32_X"},
+      /*output_descriptors*/ {"f32_X", "f32_X"}));
+  unsetenv("DISC_ENABLE_STITCH");
+  unsetenv("DISC_EXPECTED_KERNELS_IN_UT");
+}
+
 }  // namespace mlir_test
--- a/tao_compiler/mlir/disc/transforms/disc_fuse_splat_const.cc
+++ b/tao_compiler/mlir/disc/transforms/disc_fuse_splat_const.cc
@@ -1,0 +1,117 @@
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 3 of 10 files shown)


tao_compiler/mlir/disc/tests/regression/kstitch_fusion.cc
```diff
@@ -133,4 +133,19 @@ TEST(KStitchFusionGPUTest, KStitch3DF64) {
...
 }
 
-}  // namespace mlir_test
\ No newline at end of file
+// With splat const in the kStitch fusion
+TEST(KStitchFusionGPUTest, KStitchSimpleWithSplatConstF32) {
+  setenv("DISC_ENABLE_STITCH", "true", 1);
+  setenv("DISC_EXPECTED_KERNELS_IN_UT", "1", 1);
+  EXPECT_TRUE(feature_test_main(
+      /*mlir_file_path*/ c_ft_path + "kstitch_fusion_with_splat_const.mlir",
```
JamesTheZ, line 141: nit. Seems not a kstitch fusion. May rename it in the next PR?

tao_compiler/mlir/disc/transforms/disc_fuse_splat_const.cc
```diff
@@ -0,0 +1,118 @@
...
+  func.walk([&](lmhlo::ConstOp constant) {
+    if (constant.value().isSplat() &&
+        (constant->getParentOfType<lmhlo::FusionOp>() == nullptr)) {
+      worklist.push_back(constant);
+    }
+  });
+  for (auto constant : worklist) {
+    Value memref = constant.output();
+    DenseMap<lmhlo::FusionOp, SmallVector<Operation*, 4>> users;
+    for (Operation* user : memref.getUsers()) {
```
wyzero, line 65: What if there is memref.cast or reinterprete ops between  const and its consumers?

tao_compiler/mlir/disc/tests/regression/data/kstitch_fusion_with_splat_const.mlir
```diff
@@ -0,0 +1,12 @@
+module attributes {tf.versions = {bad_consumers = [], min_consumer = 0 : i32, producer = 0 : i32}} {
+  func @main(%arg0: tensor<?x?xf32>) -> (tensor<?xf32>, tensor<?xf32>) attributes {tf.entry_function = {inputs = "{{INPUTS}}", outputs = "{{OUTPUTS}}", input_placements="{{INPUT_PLACEMENTS}}", output_placements="{{OUTPUT_PLACEMENTS}}"}} {
+    %graph:2 = tf_executor.graph {
+      %1:2 = tf_executor.island wraps "tf.Const"() {value = dense<[1]> : tensor<1xi32>} : () -> tensor<1xi32>
```
JamesTheZ, line 4: Just curious about it. This const was not supported by previous fusion because the rank is 1, right? (Even it only contains one element.)

tao_compiler/mlir/disc/transforms/input_inline_fusion_pattern.cc
```diff
@@ -81,14 +81,20 @@ bool miscFuseHelper<ConstOp>(PatternRewriter& rewriter, Operation* user,
                              const SmallVector<memref::LoadOp>& load_ops,
                              LowerConfig* lower_config) {
   if (!isa<ConstOp>(producer)) return false;
-  auto memref_type =
-      cast<LmhloOp>(producer).getResultBuffer().getType().cast<MemRefType>();
-  assert(memref_type.getRank() == 0 && "only scalar ConstOp can be fused");
+  auto constant = cast<lmhlo::ConstOp>(producer);
+  auto memref_type = constant.output().getType().cast<MemRefType>();
+  bool is_splat = constant.value().isSplat();
```
wyzero, line 86: Does splat const always be scalar const values? is `[2, 3, 4] -> memref<2x3xf32>` valid as well?

JamesTheZ (approved):

tao_compiler/mlir/disc/transforms/disc_fuse_splat_const.cc
```diff
@@ -0,0 +1,118 @@
...
+// step 2, fuse them in to the consumers
+// step 3, if not used anymore, erase the original
+void DiscFuseSplatConstPass::runOnOperation() {
+  FuncOp func = getOperation();
+  auto* context = &this->getContext();
+
+  SmallVector<lmhlo::ConstOp, 4> worklist;
+  func.walk([&](lmhlo::ConstOp constant) {
+    if (constant.value().isSplat() &&
+        (constant->getParentOfType<lmhlo::FusionOp>() == nullptr)) {
```
wyzero, line 58: disc cpu stitch fusion currently can not fuse const a.t.m. So we may need to skip stitch fusion op on cpu.

tao_compiler/mlir/disc/transforms/disc_fuse_splat_const.cc
```diff
@@ -0,0 +1,118 @@
...
+      worklist.push_back(constant);
+    }
+  });
+  for (auto constant : worklist) {
+    Value memref = constant.output();
+    DenseMap<lmhlo::FusionOp, SmallVector<Operation*, 4>> users;
+    for (Operation* user : memref.getUsers()) {
+      if (user == constant.getOperation()) continue;
+      // TODO: support if users are non-fused lmhlo op
```
wyzero, line 67: nit: lmhlo_disc may be other possible consumers  as well.

tao_compiler/mlir/disc/transforms/disc_fuse_splat_const.cc
linearhit: why not for cpu stitch fusion?
existing tensorflow_ops already covers such cases (e.g. floor_mod.cc) on cpu backend, and the result is OK.
Any additional considerations on performance?

tao_compiler/mlir/disc/transforms/disc_fuse_splat_const.cc
linearhit: Then the const will currently not be fused. It is supposed to be safe anyway.

tao_compiler/mlir/disc/transforms/input_inline_fusion_pattern.cc
linearhit: it is always scalar.
`  let builders = [
    AttrBuilderWithInferredContext<(ins "ShapedType":$type,
                                        "ArrayRef<StringRef>":$values), [{
      return $_get(type.getContext(), type, values,
                   /* isSplat */(values.size() == 1));
    }]>,
  ];`

tao_compiler/mlir/disc/tests/regression/kstitch_fusion.cc
linearhit: it IS a kStitch kernel since Max is a reduce operation.

tao_compiler/mlir/disc/tests/regression/data/kstitch_fusion_with_splat_const.mlir
linearhit: rank 0 const was supported in previous fusion, rank n splat const (n > 0) was not supported.
in this case, due to the implicit broadcast and const fold, it's actually a rank n splat const.

tao_compiler/mlir/disc/transforms/disc_fuse_splat_const.cc
wyzero: Maybe add a TODO here?

PR #147, opened 2022-03-07
Add shape constraint for mhlo::DotOp.

JamesTheZ:
This helps to propagate more possible static information. The simplifier pass appears before dot-converter pass, and thus we need to support DotOp.


linearhit (approved):
LGTM

PR #149, opened 2022-03-08
[Torch] Setup env to support building with TensorRT

fortianyou:


pytorch_blade/ci_build/build_pytorch_blade.sh
```diff
@@ -41,29 +37,24 @@ function pip_install_deps() {
...
-    rm -rf build && python3 setup.py develop --cmake;
-}
-
 function ci_build() {
+    echo "DO TORCH_BLADE CI_BUILD"
     pip_install_deps
 
     if [ "$TORCH_BLADE_USE_CMAKE_BUILD" = "ON"  ]; then
-      cmake_build
+      extra_args="--cmake"
```
qiuxiafei, line 45: nit: How long are we going to maintain both build system?

fortianyou: Until bazel build system is stable enough. I should not be too long.

Orion34-lanbo (approved):
LGTM

PR #152, opened 2022-03-09
[placer][bridge] add a rule to handle expensive shape-computation-graph

wyzero:


tao/tao_bridge/passes/tao_mark_for_compilation_pass.cc
```diff
@@ -1507,6 +1510,65 @@ Status MarkForCompilationPassImpl::FindCompilationCandidates() {
...
+        !compile_time_fixed_shape_nodes[node->id()]) {
+      std::unordered_set<Node*> visitedNodeSet;
+      auto hasExpensiveProducers = [&](Node* node) {
+        std::vector<Node*> node_queue{node};
+        while (!node_queue.empty()) {
+          Node* cur = node_queue.back();
+          node_queue.pop_back();
+          if (!visitedNodeSet.insert(cur).second) continue;
+          for (auto e : cur->in_edges()) {
+            if (e->IsControlEdge()) continue;
```
linearhit, line 1555: What will happen with 'while' in control flow v1? which contains 'NextIteration' and not exactly DAG.

wyzero: `visitedNodeSet` is used for such case.

tao/tao_bridge/passes/tao_mark_for_compilation_pass.cc
```diff
@@ -1507,6 +1510,65 @@ Status MarkForCompilationPassImpl::FindCompilationCandidates() {
...
+            Node* src = e->src();
+            if (!compile_time_fixed_shape_nodes[src->id()]) continue;
+            if (src->type_string() == "Max") return true;
```
linearhit: I weakly prefer to use 'MatMul' to define 'expensive' rather than 'Max' which is more explainable.

linearhit (approved):

tao/tao_bridge/passes/tao_mark_for_compilation_pass.cc
wyzero: done

PR #164, opened 2022-03-13
Fix some errors that only occur under DEBUG mode.

JamesTheZ:


Yancey1989 (commented):
Do we need to enable the debug mode on CI?

tao_compiler/mlir/disc/transforms/lhlo_fusion.cc
```diff
@@ -147,7 +147,7 @@ class FusionPlanner {
     DenseSet<Cluster*> seen_clusters;
     for (Operation* op : op_list_) {
       Cluster* cluster = GetClusterForNode(op);
-      if (!seen_clusters.insert(cluster).second) continue;
+      if (cluster == nullptr || !seen_clusters.insert(cluster).second) continue;
```
wyzero, line 150: is null `cluster ` acutally valid?

PR #167, opened 2022-03-14
Dot batching optimization.

JamesTheZ:


tao_compiler/mlir/disc/disc_compiler.cc
```diff
@@ -242,6 +242,8 @@ LogicalResult LowerHLOToLLVM(ModuleOp m, const DISCLoweringOptions& options) {
     pm.addNestedPass<FuncOp>(disc_ral::createDiscGpuConvPaddingLegalization());
   }
 
+  pm.addNestedPass<FuncOp>(disc_ral::createDiscDotMergePass());
```
wyzero, line 245: Does this pass use the placement information? if not, maybe we can put it right before the placer pass?

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,557 @@
...
+  void runOnOperation() override;
+
+ private:
+  void dotBatchingSimplifier(FuncOp& func);
+};
+
+void DiscDotMergePass::runOnOperation() {
+  FuncOp func = getOperation();
+
+  dotBatchingSimplifier(func);
```
wyzero, line 543: signal pass failure if dotBatchingSimplifier fails?

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,557 @@
...
+    DotShape dot_shape;
+    Value lhs = op.lhs();
+    Value rhs = op.rhs();
+    // Initialize `dimension_numbers`.
+    dot_shape.dimension_numbers = op.dot_dimension_numbers();
+    auto lhs_batch_dims =
+        dot_shape.dimension_numbers.getLhsBatchingDimensions();
+    auto rhs_batch_dims =
+        dot_shape.dimension_numbers.getRhsBatchingDimensions();
+    assert(lhs_batch_dims.size() == rhs_batch_dims.size());
```
wyzero, line 176: do we need to check the actual value of lhs_batch_dims and rhs_batch_dims are same.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,557 @@
...
+  std::vector<Operation*> func_op_list;
+  DenseMap<Operation*, int64_t> func_op_to_id;
+  func_.walk([&](Operation* op) {
+    func_op_to_id.try_emplace(op, func_op_list.size());
+    func_op_list.push_back(op);
+  });
+  GraphCycles cycle_detector(func_op_list.size());
+  for (int node_id = 0; node_id < func_op_list.size(); ++node_id) {
+    Operation* op = func_op_list[node_id];
+    for (Value operand : op->getOperands()) {
```
wyzero, line 465: op having regions could also consume values outside and thus we need to take these implicit dependencies into consideration as well . please check the logic in the fusion pass to find more details.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,557 @@
...
+  // does not introduce cycle.
+
+  // Form cycle detector.
+  std::vector<Operation*> func_op_list;
+  DenseMap<Operation*, int64_t> func_op_to_id;
+  func_.walk([&](Operation* op) {
+    func_op_to_id.try_emplace(op, func_op_list.size());
+    func_op_list.push_back(op);
+  });
+  GraphCycles cycle_detector(func_op_list.size());
```
wyzero, line 462: cycle detector should only be used at basic block level. if we want to fuse dot general across basic block, then we should take region-based control flow into consideration.

BTW, dot batching may be viewed as dot fusion as well. If so, we could re-use  a log of code in the fusion pass.  The bad news is that we do fusion on buffer level a.t.m. We may re-visit this in the future.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,557 @@
...
+        analysis.getDimValue(lhs, lhs_contracting_dims[0]);
+    // Initialize `m_dim`.
+    int64_t lhs_rank = lhs.getType().cast<RankedTensorType>().getRank();
+    assert(lhs_batch_dims.size() + 2 == lhs_rank);
+    DenseSet<int64_t> lhs_batch_dims_set(lhs_batch_dims.begin(),
+                                         lhs_batch_dims.end());
+    for (int64_t i = 0; i < lhs_rank; i++) {
+      if ((lhs_batch_dims_set.find(i) == lhs_batch_dims_set.end()) &&
+          (i != lhs_contracting_dims[0])) {
+        dot_shape.m_dim = analysis.getDimValue(lhs, i);
```
wyzero, line 196: We only check dim-value-equality here. we maybe need to take symbolic dim equality into consideration as well.

BTW,  finding such candidate dots will became much easier  if we embed the dim-size-equality information into the tensor type.  we may re-visit this later.

```
%a = mhlo.dot_general(...) : tensor<?x?xf32, [symbolMAttr, symbolNAttr]>
...
%b = mhlo.dot_general(...) : tensor<?x?xf32, [symbolMAttr, symbolNAttr]>
```

tao_compiler/mlir/disc/disc_compiler.cc
JamesTheZ: It could be placed just after dot rewriter. By the way, are there side effects placer pass?

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: We plan to add the same-operand-dot-merging process just before the batching process.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: It is a good suggestion to use symbolic dim equality. Actually, there is no DimValue in some scenarios. As the planned reorganization for ShapeAnalysis is coming, I do not do the redundant checking. See the draft of ShapeAnalysisV2 here (which contains DimValue information within the symbolic-expr):https://github.com/JamesTheZ/BladeDISC/blob/8036255dbcee33847cb680b340675f2ce0d07703/tao_compiler/mlir/disc/transforms/shape_utils.h#L341

JamesTheZ: > It is a good suggestion to use symbolic dim equality. Actually, there is no DimValue in some scenarios. As the planned reorganization for ShapeAnalysis is coming, I do not do the redundant checking. See the draft of ShapeAnalysisV2 here (which contains DimValue information within the symbolic-expr):https://github.com/JamesTheZ/BladeDISC/blob/8036255dbcee33847cb680b340675f2ce0d07703/tao_compiler/mlir/disc/transforms/shape_utils.h#L341

It is not hard the traceback the DimValue based symbolic-expr. Seems not necessary to introduce extra attrs.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: We only batch dots within the same block (see the checking at L500~503). Is it safe? Otherwise it needs to get the block just wraps each dot-op and build cycle-detector one by one.

JamesTheZ: > cycle detector should only be used at basic block level. if we want to fuse dot general across basic block, then we should take region-based control flow into consideration.
> 
> BTW, dot batching may be viewed as dot fusion as well. If so, we could re-use a log of code in the fusion pass. The bad news is that we do fusion on buffer level a.t.m. We may re-visit this in the future.

It is a kind of fusion. Beside the problem that current fusion is for buffer level ops, it do not fully work for remote fusion (i.e., horizontal fusion) even with some code over there. Dot batching is remote fusion. If we plan to enable remote fusion, some common tool-funcs can be reused.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: Thanks for the comment. I will check that.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: Good idea.

tao_compiler/mlir/disc/disc_compiler.cc
wyzero: The placer pass will attach an attribute to each mhlo ops. if other passes after the placer pass add new ops, then these ops will miss placement attributes.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
wyzero: I mean we need to check the contents of lhs/rhs batch_dims are same, right?

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
wyzero: We can discuss this later. not blocking this PR. I personally think  type-based method is a more consistent way to represent shape constraint through different level of IR.

tao_compiler/mlir/disc/transforms/disc_passes.td
```diff
@@ -368,3 +368,8 @@ def LhloFusionInlinerPass : Pass<"lhlo-fusion-inliner", "FuncOp"> {
     "after its body has been lowered";
   let constructor = "createLhloFusionInlinerPass()";
 }
+
+def DiscDotMergePass : Pass<"disc-dot-merge", "FuncOp"> {
+  let summary = "Dot merge optimization";
+  let constructor = "createDiscDotMergePass()";
```
linearhit, line 374: Better to have some explanation for what is "dot merge" here.

tao_compiler/mlir/disc/transforms/shape_utils.cc
```diff
@@ -501,6 +501,22 @@ LogicalResult ShapeAnalysis::buildBlockDimValueMap(Block* block) {
       }
       mayCreateOrMapConstInt(operand);
       mayCreateOrMapConstInt(result);
+    } else if (isa<mhlo::DynamicReshapeOp, mhlo::DynamicBroadcastInDimOp>(op)) {
```
linearhit, line 504: is this related with the optimization of Gemm -> BatchGemm?

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,557 @@
...
+    mhlo::DotDimensionNumbersAttr dimension_numbers;
+    Type element_type;
+    bool operator==(const DotShape& other) const {
+      return (m_dim == other.m_dim) && (n_dim == other.n_dim) &&
+             (contracting_dim == other.contracting_dim) &&
+             (batching_dims == other.batching_dims) &&
+             (dimension_numbers == other.dimension_numbers) &&
+             (element_type == other.element_type);
+    }
+  };
```
linearhit, line 89: Is this actually the same as requiring: 
'''
(lhs().getType() == other.lhs().getType() &&
rhs().getType() == other.rhs().getType() &&
dimension_numbers() == other.dimension_numbers())
'''

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
linearhit: This should be guaranteed in previous passes, if the contents of lhs/rhs batch_dims do not equal, the dot general is illegal.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,557 @@
...
+}
+
+bool DotBatchingConverter::buildShapedDotMap() {
+  ShapeAnalysis analysis(func_);
+  if (failed(analysis.run())) {
+    LLVM_DEBUG(llvm::dbgs() << "ShapeAnalysis failes for dot merge.\n");
+    return false;
+  }
+  func_.walk([&](mhlo::DotGeneralOp op) {
+    DotShape dot_shape;
```
linearhit, line 167: I don't quite understand the necessity of DotShape struct here.
By telling if two DotShape is equal, it's actually the same logic as requiring:
```
(*analysis.getShape(lhs()) == *analysis.getShape(other.lhs()) &&
*analysis.getShape(rhs()) == *analysis.getShape(other.rhs()) &&
dot_dimension_numbers() == other.dot_dimension_numbers())
```
is is neccessary to analyze b/m/n/k in separate here?

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
linearhit: I also have similar points in offline discussion that DotMerge is better to be implemented as another FusionKind of the FusionPass. The only problem is that the 'horizontal fusion' is actually not working in the current codebase. I tried it for some other reasons a couple of weeks ago and it seems to me we do need some more fix. I didn't put much effort in investigating the root cause. But if we are clearer about the effort in fixing it, i would tend to choose to do this in the FusionPass instead of a standalone pass.

linearhit (commented):
ongoing, not finished

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: We do not only want to know the equality, but also need a key to cluster the same-shaped dots together in a map. DotShape is the key. A DotOp with a customized comparator is still not good because their hash is different. The key requires all the dims can be expressed, which is the separated b/m/n/k here. Another reason that I use the dims separated is to enable the chance to batch some dots with extra reshapes in the future. Meanwhile, the planned ShapeAnalysisV2 combines DimValue and symbol-expr together, which makes the separate comparison between DimValues and the whole shape-expression the same.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: > We can discuss this later. not blocking this PR. I personally think type-based method is a more consistent way to represent shape constraint through different level of IR.

Les discuss later. My point is that we may have some other approach that we even do not need to consider about the consistence.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: It worths an offline discussion. My current point is that the only common logic here is the cyclic detector. All the other logics are different. By introducing a new FusionKind, it introduces more codes than saving some codes. But I may missed something. Let's discuss it offline.

tao_compiler/mlir/disc/transforms/shape_utils.cc
JamesTheZ: It is not the logic about batching. The lack of the functions here makes the batching process fail in some cases due to failed dim-equal analysis. I will split it into a separate PR.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,557 @@
...
+    // Restore the graph.
+    graph_cycles->RemoveEdge(a, b);
+  }
+  return result;
+}
+
+// Arrange the insert-point of `op`'s operands in the same block. Do not deal
+// with cross-block operands.
+void arrangeOperandsInsertPointInBlock(Operation* op) {
+  for (auto operand : op->getOperands()) {
```
wyzero, line 60: needs to take care of implicit consumer inside region？

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,557 @@
...
+    auto operandOp = operand.getDefiningOp();
+    if ((operandOp != nullptr) && !operandOp->isBeforeInBlock(op)) {
```
wyzero, line 62: needs to make sure that `operandOp ` and op  are in the same block?

JamesTheZ: `isBeforeInBlock` guarantees that.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: Yes, I missed it.

tao_compiler/mlir/disc/disc_compiler.cc
JamesTheZ: Done.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: Done.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: Done.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: I left a TODO at the begging of the file about the possible plan of merging this pass into fusion-pass when horizontal fusion is fixed.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
JamesTheZ: Fixed.

tao_compiler/mlir/disc/transforms/shape_utils.cc
JamesTheZ: It has been moved to a separate PR: https://github.com/alibaba/BladeDISC/pull/179.

tao_compiler/mlir/disc/transforms/disc_passes.td
JamesTheZ: I have updated the summary. But I have no idea why it is not shown here...

tao_compiler/mlir/disc/transforms/tests/disc-dot-merge.mlir
```diff
@@ -0,0 +1,106 @@
+// RUN: disc-opt -disc-dot-merge -split-input-file %s -o - | FileCheck %s
+
+// CHECK-LABEL: func @dot_batching
+// This UT builds the tensor shape implicitly because current implement of
```
linearhit, line 4: nit: current implement -> the current implementation

tao_compiler/mlir/disc/transforms/shape_utils.h
```diff
@@ -91,87 +91,86 @@ class ValueWrapper {
...
+    state = (val == nullptr) ? INVALID : UNFOLD;
+  }
+  explicit DimValue(int64_t val) {
+    foldVal = val;
+    state = (val >= 0) ? FOLD : INVALID;
+  }
+  bool isValid() const { return state != INVALID; }
+  void dump() const {
+    if (state == FOLD) {
+      llvm::errs() << "Fold value: " << foldVal << "\n";
```
linearhit, line 110: Do we need errs for dump()?

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
linearhit: Will you do layout rewrite for two batchgemm if their b/m/n/ks are identical but has different layout?
If not, i still don't understand the necessity to analyze b/m/n/k in separate. To be discussed offline.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,610 @@
...
+      if (merged.ops.empty()) {
+        continue;
+      }
+      for (int64_t j = i + 1; j < clusters.size(); j++) {
+        auto& to_merge = clusters[j];
+        if (to_merge.ops.empty()) {
+          continue;
+        }
+        // Try merge.
+        TryMergeDotClusters(merged, to_merge, cycle_detector);
```
linearhit, line 182: what will happen if TryMergeDotClusters() returns false? 
is it expected for BuildDotClusters() to always return true?

linearhit: OK, i see the return value for both TryMergeDotClusters() and BuildDotClusters() are not used. 
better set to void to avoid redundant warnings.

tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -0,0 +1,610 @@
...
+
+  for (Block* block : blocks) {
+    // A map to help to cluster dots with same shape and dim-numbers together.
+    MergingShapeEqualMap equal_merge_shape_map;
+    if (!buildMergingShapeMap(block, analysis, equal_merge_shape_map)) {
+      continue;
+    }
+    // Find merging clusters.
+    SmallVector<DotCluster> merging_clusters;
+    BuildDotClusters<MergingShapeEqualMap>(block, equal_merge_shape_map,
```
linearhit, line 281: is it necessary for BuildDotClusters() to be a template function?

JamesTheZ: > is it necessary for BuildDotClusters() to be a template function?

This function is designed to be reused by same-operand-dot-merging. My same-operand-dot-merging implementation is here: https://github.com/JamesTheZ/BladeDISC/commit/73442372b5204f318787917371d2d0c0ebd45a8a#diff-37107805397b8e79aed36a0b338da94b5ef55757b91710e73852f579bda03f2fR148

linearhit (approved):
LGTM

PR #171, opened 2022-03-15
[Tensorflow]add c++ source code for tensorflow blade

Orion34-lanbo:
Add some of the c++ source code tensorflow blade, main pybind wrapper for onnx backend and custom ops/related tests.


tensorflow_blade/build.py
```diff
@@ -120,23 +120,15 @@ def check_init_file_miss(path, ignore_root=False):
...
-        pass
+# No need to do cc check, since pre-commit do cc check with clang-format
+def check(args):
+    with cwd(ROOT):
+        # every folder under python should contain a __init__.py file
+        # check tests dir
+        check_init_file_miss("tests", ignore_root=True)
+        execute("{}/bin/black --check --diff tests".format(args.python_dir))
+        execute("{}/bin/flake8 tests".format(args.python_dir))
+        # execute("{}/bin/mypy tests".format(args.python_dir))
```
qiuxiafei, line 131: Why is it commented out?

tensorflow_blade/build.py
```diff
@@ -205,25 +203,32 @@ def _write(line, cmd="build"):
 
             if not args.skip_trt:
                 _action_env("BLADE_WITH_TENSORRT", "1")
-                trt_home = os.environ.get("BLADE_TRT_HOME", "/usr/local/TensorRT")
-                _action_env("TENSORRT_VERSION", get_trt_version(trt_home))
-                _action_env("TENSORRT_INSTALL_PATH", trt_home)
+                trt_root = os.environ.get("TENSORRT_INSTALL_PATH", "/usr/local/TensorRT")
```
qiuxiafei, line 206: nit: how about naming the environment variables as what they're in torch-blade?

tensorflow_blade/build.py
```diff
@@ -255,7 +260,7 @@ def package_whl_with_bazel(args):
 
 def test_with_bazel(args):
     with cwd(ROOT):
-        execute("bazel test //tests/...")
+        execute("bazel test --config=cuda //tests/...")
```
qiuxiafei, line 263: Well, this worth a discussion. What we do internally is to generate all flags into `.bazelrc_gen`, to run a different configuration(maybe switching from CPU to GPU) one need to re-generate the `.bazelrc_gen` file. `bazel` commands are call without any `config`-like arguments. If we decide to use more config flags, we should generate as much as possible into `.bazelrc_gen`. 
Anyway, I've no preference.

tensorflow_blade/src/BUILD
```diff
@@ -2,11 +2,72 @@ load("@bazel_skylib//lib:selects.bzl", "selects")
...
+load("@local_config_blade_helper//:build_defs.bzl", "if_tensorrt_enabled", "if_internal")
 
 package(default_visibility = ["//visibility:public"])
 
+##############################################tf_blade ops################################################
+# Nested select is not supported by bazel, use an intermediat target to workaround.
+cc_library(
+    name = "tensorrt_ops",
+    deps = if_tensorrt_enabled([
+        "//src/custom_ops/trt_engine_op",
```
qiuxiafei, line 14: remove `src` folder or rename it into `tf_blade` or else. It appears in a target label but make no sence.

tensorflow_blade/src/BUILD
```diff
@@ -2,11 +2,72 @@ load("@bazel_skylib//lib:selects.bzl", "selects")
...
+cc_binary(
+    name = "tf_blade",
+    copts = tf_copts(),
+    linkopts = [
+        "-ldl",
+        "-lm",
+        # Tricky... to workaroudn undefined symbols _ZdlPvm. Bazel has already add -lstdc++
+        # to linkopts before object files, but it doesn't work. By adding -lstdc++ to end of
+        # linkopts, it works, it works. I've no idea why.
+        "-lstdc++",
```
qiuxiafei, line 55: This could be removed, i think.

qiuxiafei (commented):
Ongoing ...

tensorflow_blade/src/BUILD
Orion34-lanbo: The tf blade dir is reserved for python codes. Also I don't quite see the possible side effect of putting c++ code in src folder.

tensorflow_blade/src/BUILD
Orion34-lanbo: Done, it works okay without this "-lstdc++"

tensorflow_blade/build.py
```diff
@@ -120,23 +120,15 @@ def check_init_file_miss(path, ignore_root=False):
...
-            # check tests dir
-            check_init_file_miss("tests", ignore_root=True)
-            execute("{}/bin/black --check --diff tests".format(args.python_dir))
-            execute("{}/bin/flake8 tests".format(args.python_dir))
-            execute("{}/bin/mypy tests".format(args.python_dir))
-    if check_cc:
-        # TODO(xiafei.qiuxf): add cxx check.
-        pass
+# No need to do cc check, since pre-commit do cc check with clang-format
+def check(args):
```
fortianyou, line 124: Nit: Our `build.py` currently does a lot. of things. It does configurations, style checking, bazel building, packaging, and testing. I think we should deprecate it.

1. Install python3 dependencies by specifying requirements and deprecate venv
2. Configure bazel build with user bazelrc
3. Move these style checks into pre-commit scripts. 
4.  I prefer to use `setuptools` to build package wheel. Also, rename `build.py` to `setup.py`.
5. Call bazel building with custom extension in `setup.py`
6. Also call python unit test with `python -m unittest discover tests/....`(I don't think use bazel cache python3 UT is a good option before we can write dependencies correctly all the time.

Orion34-lanbo: 1. Now still using venv since we want to share the same build.py for internal version build.
2. Only .bazelrc is not enough for different build settings.
3. Doing check in pre-commit for cc is simple, but py check includes falke8/isort/mypy check, I don't konw whether pre-commit is capable of doing this.
4. We have a setup.py to build package wheel, it's not commited yet.
About 5 && 6, I still want follow the way we did in internal version, i.e. calling bazel directly. As regard to bazel cache, I have no preference now.

Orion34-lanbo: As for 3: pre-commit can do flake8/isort check, however I think mypy check should still be here, since mypy check errors cannot be automatically fixed.

tensorflow_blade/src/pybind.cpp
```diff
@@ -17,7 +17,30 @@
 
 #include "src/tf_compatible_version.h"
 
+#if BLADE_WITH_INTERNAL
```
qiuxiafei, line 20: How about wrap all internal bindings into a single header, so as to remove the repeated `#if`s and hie-related macros. Like:
```python
PYBIND11_MODULE(_tf_blade, m) {
  m.doc() = "Utils for tf blade.";
  m.def("compatible_tf_version", &compatible_tf_version);
#if BLADE_WITH_INTERNAL
  blade::internal::pybind(m);
#endif
}
```

third_party/bazel/blade_helper/blade_helper_configure.bzl
```diff
@@ -1,11 +1,20 @@
...
+load("//bazel:common.bzl", "get_python_bin", "get_host_environ")
 
 _PYTHON_BIN_PATH = "PYTHON_BIN_PATH"
+_BLADE_NEED_TENSORRT = "BLADE_WITH_TENSORRT"
+_BLADE_BUILD_INTERNAL = "BLADE_WITH_INTERNAL"
+_BLADE_BUILD_HIE = "BLADE_WITH_HIE"
 
 def _blade_helper_impl(repository_ctx):
     py_bin = get_python_bin(repository_ctx)
+    tensorrt_enabled = "True" if get_host_environ(repository_ctx, _BLADE_NEED_TENSORRT, "False").lower() in ["1", "true", "on"] else "False"
```
qiuxiafei, line 10: A helper function for these boolean flags? Don't repeat yourself.

third_party/bazel/blade_helper/build_defs.bzl.tpl
```diff
@@ -1,2 +1,17 @@
...
+        return select({"//conditions:default": x})
+    return select({"//conditions:default": []})
+
+def if_internal(x):
+    if %{IF_INTERNAL}:
+        return select({"//conditions:default": x})
+    return select({"//conditions:default": []})
+
+def if_hie_enabled(x):
+    if %{IF_HIE}:
```
qiuxiafei, line 15: Is it possible to hide all hie-related options?

third_party/bazel/blade_helper/blade_helper_configure.bzl
Orion34-lanbo: actually for template replacement, it requires string values.

third_party/bazel/blade_helper/build_defs.bzl.tpl
Orion34-lanbo: If we have a pybind.h for internal pybind wrapper, this is possible. Try fixing now..

tensorflow_blade/src/pybind.cpp
Orion34-lanbo: done

tensorflow_blade/src/BUILD
```diff
@@ -35,21 +92,27 @@ pybind_extension(
...
     linkopts = [
         "-Wl,-rpath,$$ORIGIN",
         "-Wl,-soname,_tf_blade.so",
     ],
-    copts = tf_copts(),
+    copts = if_tensorrt_enabled([
+        "-DBLADE_WITH_TENSORRT",
+    ]) + if_internal([
+        "-DBLADE_WITH_INTERNAL",
+    ]) + if_hie_enabled([
```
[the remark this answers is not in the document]
Orion34-lanbo: done

tensorflow_blade/build.py
```diff
@@ -184,46 +176,58 @@ def _write(line, cmd="build"):
         # TF-Blade
         _action_env("BLADE_WITH_TF_BLADE", "1")
 
+        if args.internal:
```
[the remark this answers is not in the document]
Orion34-lanbo: done

qiuxiafei (approved):

tensorflow_blade/src/custom_ops/non_max_suppression/gpu_non_max_suppression_op.cu.cc
```diff
@@ -0,0 +1,614 @@
...
+ * None Maximum Suppression Operation - GPU Version
+ *
+ * Author:
+ *   Yifan Lu (evanlu.lyf@alibaba-inc.com)
+ *
+ ================================================== */
+
+#ifndef __NONE_MAXIMUM_SUPPRESSION_OP_GPU_DEVICE__
+#define __NONE_MAXIMUM_SUPPRESSION_OP_GPU_DEVICE__
+
```
linearhit, line 22: Is it necessary to release these custom ops to Github?
I tend to keep them in the internal repository since: (1) the architecture is confusing to external readers; (2) I don't see strong necessity to do this.

PR #177, opened 2022-03-21
[Torch] improve TensorRT fallback & dynamic

fortianyou:
This pull request mainly improves dynamic shape optimization and fallback of TensorRT.

Previously, we only support user configuration of dynamic shapes by setting min/max/opts shapes, which are lists of ints.
And TorchBlade will automatically generate test data according to the input shapes. With the test data, TorchBlade can trace the intermediate tensors' shapes and data types which is valuable to the optimization.

1. In this PR we would like to support inferencing dynamic shape setting with user input test data. Because some network's computations could be different according to different inputs, such as detection and decoder networks. And a randomly generated test data might output empty results.

2. In this PR we improve the fallback performance of TensorRT as well.
3. In this PR we added some debug utilities.

### DIFF
--- a/pytorch_blade/src/compiler/backends/engine_class.cpp
+++ b/pytorch_blade/src/compiler/backends/engine_class.cpp
@@ -57,17 +57,21 @@ torch::List<torch::Tensor> EngineClass::Execute(
 
   torch::List<torch::Tensor> outputs;
   // do inference
-  try {
-    outputs = engine_->Execute(inputs);
-  } catch (const std::runtime_error& error) {
-    const auto& enable_fallback =
-        std::getenv("TORCH_BLADE_DISABLE_RUNTIME_FALLBACK");
-    if (enable_fallback == nullptr ||
-        std::strcmp(enable_fallback, "true") != 0 ||
-        std::strcmp(enable_fallback, "on") != 0) {
-      outputs = Fallback(inputs);
-    } else {
-      throw error;
+  if (engine_->ShouldFallback(inputs)) {
+    outputs = Fallback(inputs);
+  } else {
+    try {
+      outputs = engine_->Execute(inputs);
+    } catch (const std::runtime_error& error) {
+      const auto& enable_fallback =
+          std::getenv("TORCH_BLADE_ENABLE_RUNTIME_FALLBACK");
+      if (enable_fallback == nullptr ||
+          std::strcmp(enable_fallback, "true") == 0 ||
+          std::strcmp(enable_fallback, "on") == 0) {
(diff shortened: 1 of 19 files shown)


pytorch_blade/src/compiler/backends/engine_class.cpp
```diff
@@ -57,17 +57,21 @@ torch::List<torch::Tensor> EngineClass::Execute(
...
-      throw error;
+  if (engine_->ShouldFallback(inputs)) {
+    outputs = Fallback(inputs);
+  } else {
+    try {
+      outputs = engine_->Execute(inputs);
+    } catch (const std::runtime_error& error) {
+      const auto& enable_fallback =
+          std::getenv("TORCH_BLADE_ENABLE_RUNTIME_FALLBACK");
+      if (enable_fallback == nullptr ||
```
qiuxiafei, line 68: It's better to have this judgement as utility function or whatever.

pytorch_blade/torch_blade/onnx_backends/backend_conversion.py
```diff
@@ -57,6 +57,16 @@ def _build_onnx_engine(subgraph, engine_build_func, q_info=None, dynamic_setting
     if dynamic_settings is not None:
         dynamic_shapes = dynamic_settings
         dynamic_axes = dynamic_shapes[0].dynamic_setting
+        min_shape = dynamic_shapes[0].min_shape
+        max_shape = dynamic_shapes[0].max_shape
+        opt_shapes = dynamic_shapes[0].opt_shapes
+        logger.info(
+            "Dynamic shape settings: \n"
+            "min_shape: {str(min_shape)},\n"
```
qiuxiafei, line 65: missing the `f` prefix? btw, I usually don't log into multiple lines, feeling not good when `grep`ing.

pytorch_blade/torch_blade/tensorrt/dynamic_shapes.py
```diff
@@ -42,26 +43,39 @@ def get_shape_for_one_input(c_module, input_shapes, trt_unsupported):
     # https://github.com/pytorch/pytorch/blob/v1.8.1/torch/csrc/jit/python/python_ir.cpp#L349
     # https://github.com/pytorch/pytorch/blob/v1.8.1/torch/csrc/jit/ir/ir.cpp#L688
     # https://github.com/pytorch/pytorch/blob/v1.8.1/torch/csrc/jit/ir/ir.cpp#L658
+    def _is_list_of_int(shape):
```
qiuxiafei, line 46: I saw this function twice, in this cr ~

PR #178, opened 2022-03-21
[Torch] refactor for torch 1.11

fortianyou:


pytorch_blade/src/compiler/jit/torch/freeze_module.cpp
```diff
@@ -336,7 +336,7 @@ class AttributePropagator {
       }
     } else if (attr.isTuple()) {
       auto tuple = std::move(attr).toTuple();
-      std::vector<IValue>& elems = tuple->elements();
+      std::vector<IValue> elems = tuple->elements();
```
qiuxiafei, line 339: Why this is changed to non-ref type?

qiuxiafei (approved):

pytorch_blade/src/compiler/jit/torch/freeze_module.cpp
fortianyou: The return type of `tuple->elements()` has changed to `const &`.

PR #179, opened 2022-03-21
DimValue mapping for reshape, broadcast and slice.

JamesTheZ:


tao_compiler/mlir/disc/transforms/shape_utils.cc
```diff
@@ -449,6 +491,25 @@ LogicalResult ShapeAnalysis::buildBlockDimValueMap(Block* block) {
       }
       auto& decompose = dimValueMulDecompose_[keyDimVal];
       decompose.emplace_back(std::move(decs));
+    } else if (isa<lmhlo::DynamicBroadcastInDimOp>(*op)) {
```
linearhit, line 494: is this redundant with line 563?

tao_compiler/mlir/disc/transforms/shape_utils.cc
```diff
@@ -497,10 +556,51 @@ LogicalResult ShapeAnalysis::buildBlockDimValueMap(Block* block) {
       DimValue result(indexCast.getResult());
 
       if (failed(mapDimValueEqual(operand, result))) {
-        return WalkResult::interrupt();
+        return WalkResult::advance();
       }
       mayCreateOrMapConstInt(operand);
       mayCreateOrMapConstInt(result);
+    } else if (isa<mhlo::DynamicReshapeOp, mhlo::DynamicBroadcastInDimOp>(op)) {
```
linearhit, line 563: what are the rules to tell if an Op has to be put here,  for example, DynamicPad/DynamicIota? 
what will happen if they are not here?

linearhit (changes requested):

tao_compiler/mlir/disc/transforms/shape_utils.cc
JamesTheZ: An op that generates shape value or utilizes shape value should be placed here. ShapeAnalysisV2 does not use such logic any more, which we can discuss offline.

tao_compiler/mlir/disc/transforms/shape_utils.cc
JamesTheZ: 563 is for mhlo dialect. It is lmhlo here.

tao_compiler/mlir/disc/transforms/shape_utils.cc
JamesTheZ: If they are not here, we may miss some shape information.

linearhit (approved):
LGTM via offline discussion

PR #181, opened 2022-03-22
[Torch] fix retrieve error message segfault

fortianyou:
We met segfault when retrieving error message after build engine failed. To make our process more robust we would like to enable log message only when TORCH_BLADE_DEBUG_LOG==on

### DIFF
--- a/pytorch_blade/src/compiler/tensorrt/bridge/tensorrt_onnx_parser.cpp
+++ b/pytorch_blade/src/compiler/tensorrt/bridge/tensorrt_onnx_parser.cpp
@@ -134,13 +134,19 @@ TrtUniquePtr<nvinfer1::ICudaEngine> TensorrtOnnxParser::BuildEngine(
     }
     auto trt_engine = TrtUniquePtr<nvinfer1::ICudaEngine>(
         context.builder->buildEngineWithConfig(*context.network, *config));
-    if (trt_engine == nullptr) {
-      LOG(INFO) << "Failed to build the engine, error message are:";
+
+    // We had meet segfault when retrive error message after build engine
+    // failed. It's a issue of TensorRT.
+    // To make our process more robust we would like to enable log
+    // message only when TORCH_BLADE_DEBUG_LOG is set.
+    bool debug_log_flag = env::ReadBoolFromEnvVar("TORCH_BLADE_DEBUG_LOG");
+    if (trt_engine == nullptr && enable_debug_log) {
+      LOG(ERROR) << "Failed to build the engine, error message are:";
       auto error_recorder = context.builder->getErrorRecorder();
       auto nb_error = error_recorder->getNbErrors();
       for (int i = 0; i < nb_error; i++) {
         auto error_msg = error_recorder->getErrorDesc(i);
-        LOG(INFO) << error_msg;
+        LOG(ERROR) << error_msg;
       }
     }
     return trt_engine;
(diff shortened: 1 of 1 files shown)


pytorch_blade/src/compiler/tensorrt/bridge/tensorrt_onnx_parser.cpp
```diff
@@ -132,7 +132,15 @@ TrtUniquePtr<nvinfer1::ICudaEngine> TensorrtOnnxParser::BuildEngine(
...
-    if (trt_engine == nullptr) {
+
+    // We had meet segfault when retrive error message after build engine
+    // failed. To make our process more robust we would like to enable log
+    // message only when TORCH_BLADE_DEBUG_LOG==on
+    const char* debug_log_flag = std::getenv("TORCH_BLADE_DEBUG_LOG");
+    bool enable_debug_log = debug_log_flag != nullptr &&
+        (std::strcmp(debug_log_flag, "on") == 0 ||
+         std::strcmp(debug_log_flag, "true") == 0);
+    if (trt_engine == nullptr && enable_debug_log) {
```
linearhit, line 143: nit: if it's tricky to use tensorflow::ReadBoolFromEnvVar()
We'd better make it into some util files.

pytorch_blade/src/compiler/tensorrt/bridge/tensorrt_onnx_parser.cpp
```diff
@@ -132,7 +132,15 @@ TrtUniquePtr<nvinfer1::ICudaEngine> TensorrtOnnxParser::BuildEngine(
...
         context.builder->buildEngineWithConfig(*context.network, *config));
-    if (trt_engine == nullptr) {
+
+    // We had meet segfault when retrive error message after build engine
+    // failed. To make our process more robust we would like to enable log
+    // message only when TORCH_BLADE_DEBUG_LOG==on
+    const char* debug_log_flag = std::getenv("TORCH_BLADE_DEBUG_LOG");
+    bool enable_debug_log = debug_log_flag != nullptr &&
+        (std::strcmp(debug_log_flag, "on") == 0 ||
+         std::strcmp(debug_log_flag, "true") == 0);
```
Orion34-lanbo, line 142: Nit: perhaps add the following condition as well
```
std::strcmp(debug_log_flag, "1") == 0
```

pytorch_blade/src/compiler/tensorrt/bridge/tensorrt_onnx_parser.cpp
fortianyou: Done.

pytorch_blade/src/compiler/tensorrt/bridge/tensorrt_onnx_parser.cpp
```diff
@@ -134,13 +134,20 @@ TrtUniquePtr<nvinfer1::ICudaEngine> TensorrtOnnxParser::BuildEngine(
     }
     auto trt_engine = TrtUniquePtr<nvinfer1::ICudaEngine>(
         context.builder->buildEngineWithConfig(*context.network, *config));
-    if (trt_engine == nullptr) {
-      LOG(INFO) << "Failed to build the engine, error message are:";
+
+    // We had meet segfault when retrive error message after build engine
+    // failed. It's a issue of TensorRT.
+    // To make our process more robust we would like to enable log
```
bddppq, line 140: When `TORCH_BLADE_DEBUG_LOG` is set, the following code will still try to get error msg and run into the same segfault?

fortianyou: Yes, but occasionally. We had only run into the segfault when trying to optimize TransformerXL.

linearhit (approved):
LGTM

PR #183, opened 2022-03-22
[Torch] Add benchmark for ONNX-TensorRT

fortianyou:
Related to https://github.com/alibaba/BladeDISC/issues/157

### DIFF
--- a/.github/workflows/ngc_reusable.yml
+++ b/.github/workflows/ngc_reusable.yml
@@ -1,6 +1,8 @@
 name: NGC Reusable workflow
 
 on:
+  schedule:
+    - cron: '30 23 * * *' # 11:30 pm
   workflow_call:
     secrets:
       ALIYUN_DOCKER_USERNAME:
--- a/.github/workflows/ngc_reusable.yml
+++ b/.github/workflows/ngc_reusable.yml
@@ -1,0 +1,67 @@
+name: NGC Reusable workflow
+
+on:
+  workflow_call:
+    secrets:
+      ALIYUN_DOCKER_USERNAME:
+        required: false
+      ALIYUN_DOCKER_PASSWORD:
+        required: false
+      DOCKER_USERNAME:
+        required: false
+      DOCKER_PASSWORD:
+        required: false
+
+jobs:
+  TORCH-NGC:
+    if: github.repository == 'alibaba/BladeDISC'
+    # The type of runner that the job will run on
+    runs-on: [self-hosted, gpu-t4]
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    # Runs a single command using the runners shell
+    - name: Checkout
+      uses: actions/checkout@v2.4.0
+    - name: Build Dev Docker
+      run: |
+        set -e
+        git submodule sync
+        git submodule update --depth=1 --init --recursive
+        # docker pull bladedisc/bladedisc:latest-devel-pt-ngc
+        # docker build --cache-from bladedisc/bladedisc:latest-devel-pt-ngc -t disc-dev-pt-ngc \
+        docker build -t disc-dev-pt-ngc \
+          --build-arg BASEIMAGE=nvcr.io/nvidia/pytorch:21.11-py3 -f docker/dev/Dockerfile.torch.ngc .
+    - name: Build and Test DISC
+      run: |
+        set -e
+        mkdir -p $HOME/.cache/.torch/
+        mkdir -p $HOME/.cache/.torch/.cache
+        cp $HOME/.cache/remote_cache_token $HOME/.cache/.torch/.cache
+
+        nvidia-docker run --rm -t --user $(id -u) \
+          -v $HOME/.cache/.torch/:$HOME/.torch \
+          -v $HOME/.cache/.torch/.cache:$HOME/.cache \
+          -v /etc/passwd:/etc/passwd:ro \
+          -v /etc/group:/etc/group:ro \
+          -v $PWD:/disc \
+          -e GITHUB_WORKFLOW=$GITHUB_WORKFLOW \
+          -e TORCH_BLADE_CI_BUILD_TORCH_VERSION=ngc \
+          -w /disc/pytorch_blade \
+          disc-dev-pt-ngc bash ./benchmark/run_benchmark.sh
+    - name: Deploy PyTorch Blade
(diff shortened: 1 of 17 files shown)


.github/workflows/ngc_reusable.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Build and Test DISC
+      run: |
+        set -e
+        mkdir -p $HOME/.cache/.torch/
+        mkdir -p $HOME/.cache/.torch/.cache
+        cp $HOME/.cache/remote_cache_token $HOME/.cache/.torch/.cache
+
+        nvidia-docker run --rm -t --user $(id -u) \
+          -v $HOME/.cache/.torch/:$HOME/.torch \
+          -v $HOME/.cache/.torch/.cache:$HOME/.cache \
```
qiuxiafei, line 43: confused about above operations on `$HOME/.cache/.torch/.cache` directory.

docker/dev/Dockerfile.torch.ngc
```diff
@@ -0,0 +1,21 @@
+ARG BASEIMAGE=nvcr.io/nvidia/pytorch:21.11-py3
+FROM ${BASEIMAGE}
+
+COPY docker/scripts /install/scripts
+
+RUN mkdir -p /usr/local/TensorRT && \
+    ln -s /usr/include/x86_64-linux-gnu/ /usr/local/TensorRT/include && \
```
qiuxiafei, line 7: looks weird ...🤔

.github/workflows/ngc_reusable.yml
fortianyou: Bazel's local cache would be polluted if we have multiple versions of cuda, since we use cuda 11.5 in NGC which differs from the normal CI docker's image.

That's why we mount another cache directory.

docker/dev/Dockerfile.torch.ngc
fortianyou: NGC's TensorRT library is installed in `/lib/x86_64-linux-gnu/`. But our build scripts use `/usr/local/TensorRT` as the install path. It might be a little hack, but I think it's clean.

.github/workflows/ngc_reusable.yml
Yancey1989, line 43: Does a new cache dir increase the disc usage of the CI host?  How about only using the remote cache instead of mounting any host cache dir?

.github/workflows/ngc_reusable.yml
```diff
@@ -0,0 +1,66 @@
+name: NGC Reusable workflow
```
Yancey1989, line 1: confused about the NGC docker, can all PyTorch CI job use it as devel Docker?

pytorch_blade/ci_build/build_pytorch_blade.sh
```diff
@@ -43,7 +43,6 @@ function ci_build() {
     else
       extra_args=""
       python3 ../scripts/python/common_setup.py
-      bazel clean --expunge
```
Yancey1989, line 46: Actually don't need to clean the bazel cache ?

.github/workflows/ngc_reusable.yml
fortianyou: This NGC docker is prepared for the ONNX-TensorRT backend's benchmark.

The NGC docker is officially released by NVIDIA and keep updated monthly,  with the following libraries installed:
[CUDA](https://developer.nvidia.com/cuda-toolkit)
[cuBLAS](https://developer.nvidia.com/cublas)
[NVIDIA cuDNN](https://developer.nvidia.com/cudnn)
[NVIDIA NCCL](https://developer.nvidia.com/nccl) (optimized for [NVLink](https://www.nvidia.com/object/nvlink.html))
[RAPIDS](https://rapids.ai/)
[NVIDIA Data Loading Library (DALI)](https://developer.nvidia.com/dali)
[TensorRT](https://developer.nvidia.com/tensorrt)
[Torch-TensorRT](https://nvidia.github.io/Torch-TensorRT/)

pytorch_blade/ci_build/build_pytorch_blade.sh
fortianyou: I had moved the clean to another place.

.github/workflows/ngc_reusable.yml
fortianyou: Yes, it would increase the disk usage, but not too much. 
image

I think we can do some experiments on disc_cache and remote_cache, with a little sacrifice of disk space. If we finally come to a conclusion that remote_cache is superior to disk_cache then when just deprecate it.

.github/workflows/ngc_reusable.yml
Yancey1989: Thanks, as your introduction, I prefer to schedule the benchmark job daily or weekly, ref: https://github.com/marketplace/actions/schedule-job-action#define-scheduled-job

Yancey1989: Another but not related with this PR, how do we want to show the benchmark numbers? I found some pages with Google...
- https://labs.quansight.org/blog/2021/08/github-actions-benchmarks/
- https://github.com/marketplace/actions/continuous-benchmarking-for-go

It seems that we need a web server to vistualize these numbers...

fortianyou: Done. I added a crontab.

.github/workflows/ngc_reusable.yml
qiuxiafei: So, our cache has no knowledge about different cuda versoin? Sounds like a trouble ...

qiuxiafei: `--disc_cache` overwrites remote cache, see https://docs.bazel.build/versions/main/remote-caching.html#disk-cache . But I think the problem behind the sense is that our cache could NOT deal with multiple CUDA version correctly.

.github/workflows/ngc_reusable.yml
qiuxiafei: nit: If the file name and stage name could be renamed to benchmark-bla-bla-bla?

.github/workflows/ngc_reusable.yml
fortianyou: It's only local cache will suffer the trouble.

qiuxiafei (approved):

PR #187, opened 2022-03-23
[Torch] reopen remote-cache

fortianyou:


Orion34-lanbo (approved):

pytorch_blade/.bazelrc
```diff
@@ -1,8 +1,9 @@
 try-import ../tf_community/.bazelrc
 try-import ../tf_community/.bazelrc.user
 
+build --disk_cache=~/.cache
```
qiuxiafei, line 4: well, this is to use local disk as 'remote cache' ... so the true 'remote-cache' is still disabled IMO.

PR #192, opened 2022-03-25
GPU Conv1D support.

JamesTheZ:


tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -602,22 +602,91 @@ CudnnConvParamsKey makeConvTuningCacheKey(MemRefType<T, N>& input,
...
+  std::vector<int64_t> filter_sizes(effectiveN);
+  std::vector<int64_t> output_sizes(effectiveN);
+  int64_t padding_orig_sizes = paddings.sizes[0];
+  std::vector<int32_t> padding_data(effectiveN);
+  int64_t metadata_orig_sizes = metadata.sizes[0];
+  std::vector<int32_t> metadata_data;
+
+  bool is_1d_conv = (N == 3);
+  // input, kernel, and output layout & sizes.
+  for (int64_t layout_idx = 0; layout_idx < 3; layout_idx++) {
```
fortianyou, line 627: Is that meant we use conv2d to imitate conv1d?

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -602,22 +602,91 @@ CudnnConvParamsKey makeConvTuningCacheKey(MemRefType<T, N>& input,
...
+    std::vector<int64_t>* sizes;
+    int64_t* orig_sizes;
+    if (layout_idx == 0) {
```
fortianyou, line 630: Are these codes changed because we add conv1d? Why not retain conv2d's code the same as before?

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
JamesTheZ: Yes, you are right.

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
JamesTheZ: If it gives a Conv2D, for which N is 4, the logic is the same with previous codes. The current implementation reorganizes the previous code to support both 1D and 2D.

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -602,22 +602,91 @@ CudnnConvParamsKey makeConvTuningCacheKey(MemRefType<T, N>& input,
...
+      orig_sizes = &(output.sizes[0]);
+    }
+    int64_t metadata_idx_base = layout_idx * N;
+    int64_t spatial_dim = metadata.data[metadata_idx_base + 2];
+    for (int64_t i = 0; i < N; i++) {
+      int64_t idx = metadata_idx_base + i;
+      int64_t orig_dim = metadata.data[idx];
+      int64_t dim = orig_dim;
+      if (is_1d_conv && (orig_dim > spatial_dim)) {
+        // We will insert one spatial dim just after the existing spatial dim.
```
wyzero, line 647: nit: We usually expand the one dim right before the original spatial dim (like we do on disc-cpu or the logic in TensorFlow). Do you see any performance difference these two strategies?

JamesTheZ: The new commit inserts the extra dim just before the existing spatial dim. No performance difference for the case I tested.

wyzero (approved):
LGTM

PR #194, opened 2022-03-27
[Torch] refactor logging and debug tools

fortianyou:
1. Change logging to tf logging
2. Add read environments variables utilities
3. CMake build will be broken. To unlock the ci, we will disable TorchBlade CMake CI, since it will be deprecated soon.

### DIFF
--- a/pytorch_blade/src/common_utils/BUILD
+++ b/pytorch_blade/src/common_utils/BUILD
@@ -1,27 +1,54 @@
+package(default_visibility = ["//visibility:public"])
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
 
-filegroup(
-    name = "torch_blade_common_srcs",
-    srcs = glob(["**/*.cpp"], exclude = ["**/*_test.cpp"]),
+cc_library(
+    name = "torch_blade_logging",
+    hdrs = ["logging.h"],
+    deps = [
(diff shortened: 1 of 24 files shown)


pytorch_blade/src/common_utils/BUILD
```diff
@@ -1,27 +1,54 @@
...
 
-filegroup(
-    name = "torch_blade_common_srcs",
-    srcs = glob(["**/*.cpp"], exclude = ["**/*_test.cpp"]),
+cc_library(
+    name = "torch_blade_logging",
+    hdrs = ["logging.h"],
+    deps = [
+       "@local_org_torch//:libtorch",
+       "@org_tensorflow//tensorflow/core/platform:logging",]
```
qiuxiafei, line 9: Intend to use TF logging facilities? PyTorch-Blade depends on TF? Sounds weird.

fortianyou: Yes. I think sources dependencies are OK since we had depends on BladeDISC.

fortianyou: I think a good point is that this doesn't add new repository dependencies to TorchBlade. And the dependencies is narrow since it related to target "@org_tensorflow//tensorflow/core/platform:logging".

qiuxiafei (approved):

PR #204, opened 2022-03-29
adapt for internal blade build

Orion34-lanbo:
Make the internal code share bazel rules from open source repo

Currently works for tensorflow-gpu==2.4.0, need more test.

### DIFF
--- a/tensorflow_blade/build.py
+++ b/tensorflow_blade/build.py
@@ -182,7 +182,7 @@ def configure_with_bazel(args):
             _action_env("GCC_HOST_COMPILER_PATH", which("gcc"))
             _action_env("TF_CUDA_CLANG", "0")
             _action_env("TF_CUDA_VERSION", cuda_ver)
-            _action_env("TF_cuda_homeS", cuda_home)
+            _action_env("TF_CUDA_HOME", cuda_home)
             _action_env("TF_CUDNN_VERSION", cudnn_ver)
             if '11\.' in cuda_ver:
                 _action_env("TF_CUDA_COMPUTE_CAPABILITIES", "7.0,7.5,8.0")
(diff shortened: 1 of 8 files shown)


tensorflow_blade/build.py
```diff
@@ -182,7 +182,7 @@ def _write(line, cmd="build"):
             _action_env("GCC_HOST_COMPILER_PATH", which("gcc"))
             _action_env("TF_CUDA_CLANG", "0")
             _action_env("TF_CUDA_VERSION", cuda_ver)
-            _action_env("TF_cuda_homeS", cuda_home)
+            _action_env("TF_CUDA_HOME", cuda_home)
```
qiuxiafei, line 185: 😂

third_party/bazel/tf/tf_configure.bzl
```diff
@@ -23,21 +37,27 @@ def _tf_configure_impl(repository_ctx):
 
     tf_major = repository_ctx.os.environ[_TF_MAJOR_VERSION]
     tf_minor = repository_ctx.os.environ[_TF_MINOR_VERSION]
+    tf_is_pai = repository_ctx.os.environ[_TF_IS_PAI].lower() in ["1", "true", "on"]
```
qiuxiafei, line 40: Don't repeat this boolean check logic.

third_party/bazel/cuda_static/cuda_static_configure.bzl
```diff
@@ -0,0 +1,20 @@
+load("//bazel:common.bzl", "files_exist", "get_bash_bin")
+
+_TF_CUDA_HOME = "TF_CUDA_HOME"
```
qiuxiafei, line 3: A little bit worry about inconsistence on env var prefix `TF_`. This file is expect to be a common tool, if torch_blade reference it and has to write a `TF_`-prefixed env var, which is counterintuitive.

Orion34-lanbo: As I know, torch blade also uses the bazel rules for gpus/cuda etc from tf community, I think this may not be a problem now ...

third_party/bazel/tf/tf_configure.bzl
Orion34-lanbo: fixed

qiuxiafei (approved):

PR #211, opened 2022-03-31
[CI][TensorFlow] make bazel rules compatible with different settings

Orion34-lanbo:
1. add rule for `libiomp5.so`
2. depend `@local_config_cuda//cuda:cublasLt` only when `if_has_cublaslt` is true make it works for cu100 or cpu device
3. link myelin if myelin libs exist


tensorflow_blade/src/tensorrt/BUILD
```diff
@@ -57,6 +58,9 @@ cc_library(
         "//src/util:logging",
     ] + if_has_cublaslt([
         "@local_config_cuda//cuda:cublasLt",
+    ]) + if_has_myelin([
+        "@local_config_cuda_static//:nvrtc",
```
qiuxiafei, line 62: why is nvrtc bind with myelin ?

third_party/bazel/cuda_static/cuda_static.BUILD.tpl
```diff
@@ -21,3 +21,8 @@ cc_import(
     name = "cudnn_static",
     static_library = "lib64/libcudnn_static.a",
 )
+
+cc_import(
+    name = "nvrtc",
+    shared_library = "lib64/libnvrtc.so",
```
qiuxiafei, line 27: an shared lib in a build file named static...

third_party/bazel/tensorrt/repo.bzl
```diff
@@ -1,14 +1,66 @@
...
+    has_myelin = files_exist(
+            repo_ctx,
+            [
+                "lib/libmyelin_compiler_static.a",
+                "lib/libmyelin_executor_static.a",
+                "lib/libmyelin_pattern_library_static.a",
+                "lib/libmyelin_pattern_runtime_static.a",
+            ]
+    )
+    if_has_myelin = has_myelin[0] and has_myelin[1] and has_myelin[2] and has_myelin[3]
```
qiuxiafei, line 57: pls use `all`: https://docs.bazel.build/versions/main/skylark/lib/globals.html#all

third_party/bazel/tf/tf_configure.bzl
```diff
@@ -44,15 +44,15 @@ def _tf_configure_impl(repository_ctx):
         "-DTF_MAJOR={}".format(tf_major),
         "-DTF_MINOR={}".format(tf_minor),
     ]
-    if tf_is_pai:
+    if tf_is_pai == "True":
```
qiuxiafei, line 47: looks weird ... why not making `tf_is_pai` a boolean variable, and use `str(tf_is_pai)` for templating.

third_party/bazel/cuda_static/cuda_static.BUILD.tpl
Orion34-lanbo: I did not find a better for this supplementary cuda rule, may cuda_supplement?

tensorflow_blade/src/tensorrt/BUILD
Orion34-lanbo: For trt7, if nvrtc is not linked when myelin libs are linked, we will encounter undefined symbol issue, this symbol is from nvrtc. However for later trt versions, such as trt8, myelin libs no longer exists, we don't need to link nvrtc

third_party/bazel/tensorrt/repo.bzl
Orion34-lanbo: Originally, I used `all`, another bug makes me think `all` doesn't work as expected.

third_party/bazel/tf/tf_configure.bzl
Orion34-lanbo: `tf_is_pai` is acquired from env as string use `get_env_bool_value_str` func, maybe using `if bool(tf_is_pai)` here is ok.

third_party/bazel/cuda_static/cuda_static.BUILD.tpl
Orion34-lanbo: done

third_party/bazel/tensorrt/repo.bzl
Orion34-lanbo: done

third_party/bazel/tf/tf_configure.bzl
Orion34-lanbo: done

qiuxiafei (approved):

Issue #220, opened 2022-04-06
Enhance ONNX clustering for pytorch_blade


qiuxiafei:
1. Add support for `aten::upsample_bilinear2d` TensorRT backend, with opset 11 specified.
2. Try best to support subgraph with scalar input.

PR #221, opened 2022-04-06
[to #220] Enhance ONNX clustering for pytorch_blade.

qiuxiafei:


pytorch_blade/torch_blade/onnx_backends/backend_testbed.py
```diff
@@ -387,21 +410,26 @@ def _appendNode(self, old_node):
...
             for inp in node.inputs():
                 all_used_values.add(inp)
             for out in node.outputs():
-                all_inter_outputs.add(out)
+                if out in self._segment2orig_value_map:
+                    orig_out = self._segment2orig_value_map[out]
+                    if orig_out.uses():
+                        all_inter_outputs.add(out)
+                else:
+                    all_inter_outputs.add(out)
```
fortianyou, line 432: Are these modified to filter out the unused value from per-original node subgraph outputs?

I think filtering out the subgraph outputs from `maybe_graph_output` at L434 would be more clear:
```python
subgraph_outputs = set(out for out in maybe_graph_output if out used in original graph)
```

pytorch_blade/torch_blade/pass_manager.py
```diff
@@ -61,7 +75,7 @@ def _export_onnx(graph, dynamic_axes):
     custom_opsets = {}
     val_add_node_names = True
     proto = graph._export_onnx(
-        dict(),
+        params_dict,
```
fortianyou, line 78: Will this freeze extra-parameters into the proto as constants?

pytorch_blade/torch_blade/onnx_backends/backend_testbed.py
```diff
@@ -38,17 +38,37 @@ def is_tensor(val):
...
+                scalar_inputs.append(None)
+        return scalar_inputs
+
     def __call__(self):
         try:
             graph = self._graph
             # use mannual rules to filter the graph
             if not onnx_lower_guard.check_graph_with_rules(graph):
                 return False
+            scalar_types = self._record_inputs_scalar_types(graph)
```
fortianyou, line 65: Why should these scalar exists, since we had forbidden scalar as subgraph inputs without ingore_device=True, as is L379.

pytorch_blade/src/compiler/jit/fusion.cpp
```diff
@@ -125,5 +125,16 @@ torch::TypePtr get_list_tensor_type() {
   return list_type;
 }
 
+torch::TypePtr tensor_type_from_scalar(const c10::Type& typ) {
```
fortianyou, line 128: Nit: I prefer `create_tensor_type_from_scalar_type`，also it seems duplicate to https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/core/jit_type.h#L587

pytorch_blade/torch_blade/onnx_backends/backend_testbed.py
fortianyou: I think we can put some declarations and examples here.

pytorch_blade/torch_blade/onnx_backends/backend_testbed.py
qiuxiafei: refactored.

pytorch_blade/torch_blade/pass_manager.py
qiuxiafei: No. `params_dict` should contain frozen constants only. `_jit_pass_onnx_constfold` will return a `params_dict` like that.

pytorch_blade/torch_blade/onnx_backends/backend_testbed.py
qiuxiafei: They're indirect inputs introduced by recursive `_appendNode` and `_add_graph_input_if_need` invocations, as documented in comments update just now ~

pytorch_blade/src/compiler/jit/fusion.cpp
qiuxiafei: Renamed ~ 
Actually this piece of code taken from: 
except return nullptr on failure.

https://github.com/pytorch/pytorch/blob/75ad6fea66bc2f0ff0f58d83a1b198549034c069/aten/src/ATen/core/jit_type.h#L1575-L1586

fortianyou (approved):
Overall LGTM

PR #225, opened 2022-04-07
[AArch64] make sure PIC is used for jit compilation

wyzero:
#227  To support JIT compilation on AArch64.  The default flag on AArch64 is static not PIC.

### DIFF
--- a/tao_compiler/mlir/disc/disc_compiler.cc
+++ b/tao_compiler/mlir/disc/disc_compiler.cc
@@ -128,17 +128,36 @@ LogicalResult RewriteLLVMModule(llvm::Module* m) {
 }
 
 std::unique_ptr<llvm::TargetMachine> GetTargetMachine(llvm::Module* module) {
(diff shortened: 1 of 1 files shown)


tao_compiler/mlir/disc/disc_compiler.cc
```diff
@@ -621,6 +640,12 @@ LogicalResult LowerLLVMToBinary(ModuleOp module,
     return failure();
   }
 
+  if (tm->getRelocationModel() != llvm::Reloc::Model::PIC_) {
+    llvm::errs() << "-fPIC must be specified\n";
```
linearhit, line 644: This should never happen, the codes just functions as assert(), am i getting it correctly?

linearhit (approved):
LGTM

tao_compiler/mlir/disc/disc_compiler.cc
wyzero: Yes, it should not happen after this PR. still put this check here for double check.

PR #228, opened 2022-04-07
[bugfix] remove `_output_shapes` attr to avoid crash when loading graphdef to mlir

wyzero:
`tao_compiler_main` failed to load graphdef with ops having `_output_shapes`. Just remote such attribute here as a workaround to fix such problem.

### DIFF
--- a/tao/tao_bridge/passes/tao_encapsulate_subgraphs_pass.cc
+++ b/tao/tao_bridge/passes/tao_encapsulate_subgraphs_pass.cc
@@ -792,14 +792,14 @@ Status Encapsulator::Subgraph::RecordArg(
     DataType dtype = edge->dst()->input_type(edge->dst_input());
     builder.Attr("T", dtype);
     builder.Attr("index", arg_index);
-    if (src_node->attrs().Find("_output_shapes") != nullptr) {
-      const auto& output_shape =
-          src_node->attrs().Find("_output_shapes")->shape();
-      builder.Attr("_output_shapes", PartialTensorShape(output_shape));
-      VLOG(2) << "Adding _output_shapes info to node: "
-              << absl::StrCat(src_node->name(), "_", src_slot, "_arg");
-      VLOG(2) << "Shape info: " << output_shape.DebugString();
-    }
+    // if (src_node->attrs().Find("_output_shapes") != nullptr) {
+    //   const auto& output_shape =
+    //       src_node->attrs().Find("_output_shapes")->shape();
+    //   builder.Attr("_output_shapes", PartialTensorShape(output_shape));
+    //   VLOG(2) << "Adding _output_shapes info to node: "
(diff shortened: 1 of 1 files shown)


tao/tao_bridge/passes/tao_encapsulate_subgraphs_pass.cc
```diff
@@ -792,14 +792,18 @@ Status Encapsulator::Subgraph::RecordArg(
...
+    // such problem.
+    //
+    // if (src_node->attrs().Find("_output_shapes") != nullptr) {
+    //   const auto& output_shape =
+    //       src_node->attrs().Find("_output_shapes")->shape();
+    //   builder.Attr("_output_shapes", PartialTensorShape(output_shape));
+    //   VLOG(2) << "Adding _output_shapes info to node: "
+    //           << absl::StrCat(src_node->name(), "_", src_slot, "_arg");
+    //   VLOG(2) << "Shape info: " << output_shape.DebugString();
+    // }
```
linearhit, line 806: Just curious, what's the reason behind for tao_compiler_main failed in loading a graphdef with "_output_shapes"?

tao/tao_bridge/passes/tao_encapsulate_subgraphs_pass.cc
```diff
@@ -792,14 +792,18 @@ Status Encapsulator::Subgraph::RecordArg(
...
-    if (src_node->attrs().Find("_output_shapes") != nullptr) {
-      const auto& output_shape =
-          src_node->attrs().Find("_output_shapes")->shape();
-      builder.Attr("_output_shapes", PartialTensorShape(output_shape));
-      VLOG(2) << "Adding _output_shapes info to node: "
-              << absl::StrCat(src_node->name(), "_", src_slot, "_arg");
-      VLOG(2) << "Shape info: " << output_shape.DebugString();
-    }
+    // `tao_compiler_main` failed to load graphdef with ops having
+    // `_output_shapes`. Just remote such attribute here as a workaround to fix
```
linearhit, line 796: remote -> remove

tao/tao_bridge/passes/tao_encapsulate_subgraphs_pass.cc
wyzero: I think it's because some `_output_shapes` attributes are not correct. 

In my case, there are many ops have empty `'_output_shapes' : {}`  while the ranks of such ops are not zero, resulting in failing to load the graphdef.

The problem is there isn't an elegant way to tell if an empty `_output_shapes` is valid before we know the actual rank of the corresponding tensor.

tao/tao_bridge/passes/tao_encapsulate_subgraphs_pass.cc
wyzero: Thanks, done.

linearhit (approved):
LGTM

Issue #227, opened 2022-04-07
[disc-cpu][AArch64] ARM cpu support


wyzero:
We will start to support AArch64 CPU and use this PR to track the status.

PR #229, opened 2022-04-07
[AArch64] refator build script to build for different cpus

wyzero:


pytorch_blade/.bazelrc
```diff
@@ -3,6 +3,7 @@ try-import ../tf_community/.bazelrc.user
 
 build --disk_cache=~/.cache
 build:torch_cuda --config=cuda --config=disc_cuda --define enable_cuda=true
-build:torch_cpu --config=disc_cpu --config=disc_mkldnn
+build:torch_x86 --config=disc_x86 --config=disc_mkldnn
```
qiuxiafei, line 6: what if it's targeting arm CPU with a cuda GPU like jetson? Should both `torch_cuda` and `torch_aarch64` be specified?  While for current situations, we just specify `torch_cuda` for X86 CPU with a cuda GPU.

Or, we just will never do cross compiling? There flags are for accelerators only, host code depends on compiling platform.

scripts/python/common_setup.py
```diff
@@ -251,33 +251,67 @@ def mkldnn_build_dir(root=None):
...
+              curl -fsSL https://hlomodule.oss-cn-zhangjiakou.aliyuncs.com/mkl_package/mkl-static-2022.0.1-intel_117.tar.bz2  | tar xjv
+              curl -fsSL https://hlomodule.oss-cn-zhangjiakou.aliyuncs.com/mkl_package/mkl-include-2022.0.1-h8d4b97c_803.tar.bz2 | tar xjv
+            """
+            execute(download_cmd)
+
+    if args.aarch64:
+        with cwd(acl_dir):
+            # downlaod and build acl for onednn
+            cmd = '''
+              readonly ACL_REPO="https://github.com/ARM-software/ComputeLibrary.git"
```
qiuxiafei, line 278: Shall we just built ACL into docker image? So does to MKL ...

scripts/python/tao_build.py
```diff
@@ -155,6 +157,11 @@ def configure_compiler(root, args):
                 f.write("\n")
                 f.write("build --remote_cache={}\n".format(token))
                 f.write("test --remote_cache={}\n".format(token))
+        with open(".tf_configure.bazelrc", "a") as f:
+            f.write("\n")
+            bazel_startup_opts = "--host_jvm_args=-Djdk.http.auth.tunneling.disabledSchemes="
```
qiuxiafei, line 162: Will this work without proxy environment variables?

tao/CMakeLists.txt
```diff
@@ -86,12 +86,21 @@ find_package(Protobuf ${Tensorflow_PB_VERSION}
     QUIET
 )
 if(NOT DEFINED Protobuf_PROTOC_EXECUTABLE)
-    tao_execute_cmd('
-        DESC "download protoc"
-        PWD "."
-        COMMAND bash ${tao_ops_SOURCE_DIR}/tools/download_protoc.sh ${CMAKE_CURRENT_BINARY_DIR} ${Tensorflow_PB_VERSION}
-        STD_OUT Protobuf_PROTOC_EXECUTABLE
-    )
+    if(${TAO_AARCH64})
```
qiuxiafei, line 89: Just ensure that these will get eliminated once we've transferred to Bazel and protoc are compiled from source. @Orion34-lanbo

tao/tao_bridge/tf_compatible.h
```diff
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include "tensorflow/core/framework/resource_mgr.h"
```
qiuxiafei, line 14: Just curious why to add this include.

wyzero: we encounter a failure on arm without this header. I do not know why it's ok on x86. Maybe compiler version is different?

scripts/python/common_setup.py
wyzero: we can not pre-build this because currently oneDNN  repo relies on building ACL from source. And we do not want to use a customized oneDNN version.

scripts/python/tao_build.py
wyzero: Yes.

qiuxiafei (approved):

wyzero:
to #227

PR #232, opened 2022-04-08
[Torch] standalone pure c/c++ libtorch_blade.so

fortianyou:
RT

### DIFF
--- a/pytorch_blade/bazel_build.py
+++ b/pytorch_blade/bazel_build.py
@@ -42,6 +42,8 @@ class BazelBuild(TorchBladeBuild):
         self.test_suite = "//src:torch_blade_gtests"
         self.targets = [
             "@org_tensorflow//tensorflow/compiler/mlir/disc:disc_compiler_main",
+            "@org_tensorflow//tensorflow/compiler/mlir/xla/ral:libral_base_context.so",
+            "//src:libtorch_blade.so",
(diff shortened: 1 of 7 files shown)


pytorch_blade/bazel_build.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, *args, **kwargs):
         self.test_suite = "//src:torch_blade_gtests"
         self.targets = [
             "@org_tensorflow//tensorflow/compiler/mlir/disc:disc_compiler_main",
+            "@org_tensorflow//tensorflow/compiler/mlir/xla/ral:libral_base_context.so",
+            "//src:libtorch_blade.so",
```
qiuxiafei, line 46: nit: Target's from `@org_tensorflow` could be omitted if they were depended by targets of TorchBlade.

pytorch_blade/src/BUILD
```diff
@@ -64,9 +76,9 @@ pybind_library(
         "//visibility:public",
     ],
     deps = [
-        ":torch_blade",
+        ":torch_blade_shared",
```
qiuxiafei, line 79: Just curious if this `pybind_library` could depend on target `libtorch_blade.so` directly?

qiuxiafei (approved):

PR #243, opened 2022-04-13
Fix a bug for ral_batch_gemm when m = 1

linearhit:


tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -173,14 +173,14 @@ static bool DoGemmWithAlgorithm(
...
-      std::is_same<AlphaBeta, float>::value && m == 1) {
+      std::is_same<AlphaBeta, float>::value && m == 1 && batch_size == 1) {
     return TrySgemvInternal<InT, OutT, AlphaBeta>(
         stream, rhs_transpose, n, k,
         /*alpha=*/alpha, rhs_data,
         /*leading dim of RHS=*/rhs_matrix.num_cols, lhs_data,
         /*incx*/ 1, /*beta=*/beta, &output_data, /*incy*/ 1);
   }
-  if (algorithm) {
+  if (algorithm && batch_size == 1) {
```
wyzero, line 183: Do we need to check batch_size == 1 in if statement?  The batch_size is supposed to be 1 if algorithm is not null (just as the check in line 185).

linearhit: it's not strictly necessary here, just to prevent wrong use of DoGemmWithAlgorithm()

wyzero: The check in line 185 could also be used to protect the unintentional uses?

wyzero (approved):

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
linearhit: resolved.

PR #247, opened 2022-04-13
Fix a bug in tf.SigmoidGradOp legalize pattern

linearhit:


tao_compiler/mlir/disc/tests/tensorflow_ops/sigmoid_grad.cc
```diff
@@ -0,0 +1,61 @@
...
+#include "tensorflow/compiler/mlir/disc/tests/mlir_test.h"
+#include "tensorflow/core/platform/test.h"
+
+namespace mlir_test {
+
+const std::string c_ft_path =
+    "tensorflow/compiler/mlir/disc/tests/tensorflow_ops/data/";
+
+// static shape test case
+TEST(TFSigmoidGradOpTest, StaticShape2DF32) {
```
wyzero, line 26: Do we need to add any pattern for dynamic-shape softmaxGrad? I only see you disable some patterns in tf_community.

linearhit: we already have

wyzero (approved):
LGTM

PR #249, opened 2022-04-14
bazel build for x86/aarch64 tao_bridge

Orion34-lanbo:
Since `ACL` is compiled by `SCons`, there exists no bazel rules to call `SCons` build. I have to do it with a `sh_binary`, however the compiled results are not able to directly used by `BUILD`. Some shell actions have been inserted between several `bazel build` calls.


tensorflow_blade/build.py
```diff
@@ -172,6 +173,9 @@ def _write(line, cmd="build"):
         _opt("cxxopt", "-DBUILD_WITH_BAZEL")
         _action_env("BUILD_WITH_BAZEL", "1")
         _action_env("PYTHON_BIN_PATH", which("python3"))
+        _action_env("GCC_HOST_COMPILER_PATH", which("gcc"))
```
qiuxiafei, line 176: why it's moved to common part?

tensorflow_blade/build.py
```diff
@@ -251,10 +255,22 @@ def _write(line, cmd="build"):
             _write("--//:device=cpu")
```
qiuxiafei, line 255: `aarch64` seems to be different from `cpu`, `cpu` here means x86 i think.

tensorflow_blade/workspace0.bzl
```diff
@@ -50,6 +51,48 @@ def _tf_blade_repositories():
         register_preinstalled_tools = True,  # just use the pre-installed.
     )
 
+    maybe_http_archive(
+        name = "rules_java",
+        sha256 = "f5a3e477e579231fca27bf202bb0e8fbe4fc6339d63b38ccb87c2760b533d1c3",
+        strip_prefix = "rules_java-981f06c3d2bd10225e85209904090eb7b5fb26bd",
+        urls = ["http://pai-blade.oss-cn-zhangjiakou.aliyuncs.com/bazelbuild/rules_java/archive/981f06c3d2bd10225e85209904090eb7b5fb26bd.tar.gz"],
```
qiuxiafei, line 58: what requires java ?

Orion34-lanbo: This is tmp for build on ecs-arm servers, cannot download `rule_java` from github urls.

tensorflow_blade/build.py
Orion34-lanbo: I think these names should be unified for disc/tf_blade/torch_blade，maybe discuss offline?

tensorflow_blade/build.py
Orion34-lanbo: I think this needed for all devices build, thus move to `common` part.

Orion34-lanbo:
Latest version no longer need `hacks` for build `acl`, bazel external rules for `make` and `cmake` is applied for `acl` and `onednn` build, and they can be used as bazel `labels` in ral's BUILD.
@qiuxiafei pls review the latest changes. Thanks!
As for the replacement for `cmake` build call in `scripts/python/tao_build.py`, I will do it in the next pr.

third_party/bazel/onednn/onednn_configure.bzl
```diff
@@ -0,0 +1,28 @@
...
+_BUILD_WITH_MKLDNN = "BUILD_WITH_MKLDNN"
+_IF_CXX11_ABI = "IF_CXX11_ABI"
+
+def _onednn_impl(repository_ctx):
+    with_mkldnn = get_env_bool_value(repository_ctx, _BUILD_WITH_MKLDNN)
+    if with_mkldnn:
+        if_cxx11_abi = get_env_bool_value(repository_ctx, _IF_CXX11_ABI)
+        with_aarch64 = get_env_bool_value(repository_ctx, _BUILD_WITH_AARCH64)
+        if with_aarch64:
+            acl_root = repository_ctx.path(Label("@acl_compute_library//:Makefile")).dirname
```
Orion34-lanbo, line 13: getting `acl_root` is a little bit hacky, since this `template` action happens when loading `WORKSPACE` content. At this time, all the build output from `@acl_compute_library` are not ready yet, trying to get location of `@acl_compute_library//:build/libarm_compute.a` will not work since `libarm_compute.a` is not exist. This `Makefile` exist after `@acl_compute_library`'s download and patch is complete, thus hack like this.

scripts/python/tao_build.py
```diff
@@ -650,6 +650,12 @@ def parse_args():
         action="store_true",
         help="Build tao with cpu support only",
     )
+    parser.add_argument(
+        "--aarch64",
```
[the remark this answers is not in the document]
Orion34-lanbo: Do we have plans for `cross compilation`? @wyzero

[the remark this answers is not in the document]
qiuxiafei: Had some offline dicussion with @wyzero, `No` to keep simple.

tensorflow_blade/build.py
```diff
@@ -272,7 +285,10 @@ def _write(line, cmd="build"):
 
 def build_with_bazel(args):
     with cwd(ROOT):
-        execute("bazel build //src:_tf_blade.so")
+        if args.device == "cpu":
+            execute("bazel build @org_tao_bridge//:libtao_ops.so")
```
qiuxiafei, line 289: why not build tf_blade.so for all?

tao_compiler/mlir/xla/ral/BUILD
```diff
@@ -477,22 +477,65 @@ tf_gpu_kernel_library(
...
+)
+
 cc_library(
     name = "common_context_bridge",
     srcs = [
         "context/common_context_impl.cc",
-    ] + if_cuda_or_rocm([
         "context/common_context_impl_cuda.cc",
         "context/stream_executor_based_impl.cc",
+    ] + if_cuda_or_rocm([
```
qiuxiafei, line 529: why `context/common_context_impl_cuda.cc` and `context/stream_executor_based_impl.cc` are moved out of `if_cuda_or_rocm`

tensorflow_blade/workspace0.bzl
```diff
@@ -50,6 +53,51 @@ def _tf_blade_repositories():
...
+        build_file = "@org_third_party//bazel/mkldnn:mkl_include.BUILD",
+        sha256 = "3df729b9fa66f2e1e566c70baa6799b15c9d0e5d3890b9bd084e02299af25002",
+        urls = [
+            "https://hlomodule.oss-cn-zhangjiakou.aliyuncs.com/mkl_package/mkl-include-2022.0.1-h8d4b97c_803.tar.bz2",
+            "https://hlomodule.oss-cn-zhangjiakou.aliyuncs.com/mkl_package/mkl-include-2022.0.1-h8d4b97c_803.tar.bz2",
+        ],
+    )
+
+    # for aarch64 related acl library
+    tf_http_archive(
```
qiuxiafei, line 85: No strong opinion but I prefer to depend on tf as less as possible. this a normal `http_archive` be here?

third_party/bazel/acl/acl_makefile.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/Makefile b/Makefile
+new file mode 100644
+index 0000000..bd9a183
+--- /dev/null
++++ b/Makefile
+@@ -0,0 +1,10 @@
++all:
```
qiuxiafei, line 7: A patch works!
But is it possible to use `build_data` of `make` rule?

tao_compiler/mlir/xla/ral/BUILD
```diff
@@ -477,22 +477,65 @@ tf_gpu_kernel_library(
...
+        "-L$(location @mkl_static//:lib)/",
+        "-Wl,--start-group",
+        "$(location @mkl_static//:lib)/libmkl_intel_ilp64.a",
+        "$(location @mkl_static//:lib)/libmkl_gnu_thread.a",
+        "$(location @mkl_static//:lib)/libmkl_core.a",
+        "-Wl,--end-group",
+    ]) + if_disc_aarch64([
+        "-L$$(dirname $(location @acl_compute_library//:build/libarm_compute.a)",
+        "-Wl,--start-group",
+        "$(location @acl_compute_library//:build/libarm_compute.a)",
```
qiuxiafei, line 502: Is it OK to explicitly depend on `@acl_compute_library//:acl` instead of writing such `-L`s and `.a`s? An example: https://github.com/bazelbuild/rules_foreign_cc/blob/main/examples/make_simple/BUILD.bazel#L32

third_party/bazel/blade_disc_helper/BUILD.tpl
```diff
@@ -0,0 +1,33 @@
+package(default_visibility = ["//visibility:public"])
+
+config_setting(
```
qiuxiafei, line 3: lol, we need a `global_helper`

third_party/bazel/blade_disc_helper/blade_disc_helper_configure.bzl
```diff
@@ -34,6 +34,7 @@ def _blade_disc_helper_impl(repository_ctx):
         "%{DISC_BUILD_HOST}": get_host_environ(repository_ctx, _DISC_BUILD_HOST, ""),
         "%{DISC_BUILD_IP}": get_host_environ(repository_ctx, _DISC_BUILD_IP, ""),
         "%{DISC_BUILD_TIME}": get_host_environ(repository_ctx, _DISC_BUILD_TIME, ""),
+	"%{IF_CXX11_ABI}": "False",
```
qiuxiafei, line 37: bad indent...

third_party/bazel/mkldnn/mkl_include.BUILD
```diff
@@ -0,0 +1,3 @@
+package(default_visibility = ["//visibility:public"])
+
+exports_files(["include"])
```
qiuxiafei, line 3: is it possible to expose those as header-only targets instead of files? So does static libs.

third_party/bazel/third_party/BUILD
```diff
@@ -0,0 +1,3 @@
+package(default_visibility = ["//visibility:public"])
+
+exports_files(["compute_library.sh"])
```
qiuxiafei, line 3: export for what ?

third_party/bazel/onednn/onednn_configure.bzl
qiuxiafei: pls put your comment into source code as comments.

qiuxiafei:
Genrally looks great to me. Pls take a look at new, but trivial, comments ~

tao_compiler/mlir/xla/ral/BUILD
Orion34-lanbo: Here we try to get the location of static libraries from `@acl_compute_library`. However the target `@acl_compute_library//:acl` has 5 outputs as below, getting location will fail since bazel doesn't know which target's location to get. The example from `make_simple` works because it is used as `deps`, here we need the specific location.
`bazel-bin/external/acl_compute_library/acl
bazel-bin/external/acl_compute_library/acl/build/libarm_compute.a
bazel-bin/external/acl_compute_library/acl/build/libarm_compute_core.a
bazel-bin/external/acl_compute_library/acl/build/libarm_compute_graph.a
bazel-bin/external/acl_compute_library/copy_acl
`

third_party/bazel/acl/acl_makefile.patch
Orion34-lanbo: I think it will not work, I tried adding it in `build_data`, however when `make` is called, the `Makefile` cannot be found under the acl's root code dir.

tensorflow_blade/workspace0.bzl
Orion34-lanbo: I think it may need some extra work here, since the built-in `patch` mechanism in bazel doesn't work for the current patch format.

tao_compiler/mlir/xla/ral/BUILD
Orion34-lanbo: These modification follows how `tao_bridge` works in `CMake`. And if not removed out of `if_cuda_or_rocm `, build for cpu device will failed. Also these two files doesn't contain `cuda_or_rocm` device related code, just the wrapping cpp code.

tensorflow_blade/build.py
Orion34-lanbo: Currently `tf_blade` has no cpu related code ..., just a placeholder here.

third_party/bazel/third_party/BUILD
Orion34-lanbo: useless now, will be removed.

tao_compiler/mlir/xla/ral/BUILD
qiuxiafei: sorry, i miss the `-Wl` flag..

tao_compiler/mlir/xla/ral/BUILD
Orion34-lanbo: After adding related flags, this is no longer needed.

tensorflow_blade/build.py
Orion34-lanbo: Modified to build `//src:_tf_blade.so` with different configs

qiuxiafei (approved):

PR #258, opened 2022-04-15
Fix a bug in disc-dot-merger pass

linearhit:


tao_compiler/mlir/disc/transforms/disc_dot_merge.cc
```diff
@@ -357,6 +357,7 @@ bool DotShareOperandMergeConverter::applyMerging(DotCluster& cluster,
     auto concat_dim_size = concat_op_type.getDimSize(concat_dim);
     if (concat_dim_size == ShapedType::kDynamicSize) {
       is_dynamic_shape = true;
+      concat_dim_sum = ShapedType::kDynamicSize;
```
JamesTheZ, line 360: Looks like we do not need to deal with the dynamic case for `concat_dim_sum`. According to L368 and L372, if `concat_dim_sum` is `true`, for the concat-dim, the value in `concat_op_shapes` will not be updated, remaining `kDynamicSize`.
(The logic for L372~374 is confusing and a better way is to move the assignment of concat-dim outside the loop.)

linearhit: the init value of concat_dim_sum is zero, resulting in unexpected zero dim size in case  the concat_dim_size is -1

JamesTheZ: I see. The error use is at L395. Your fix is correct. By the way, it looks like just removing L395 directly fixes this problem too.

JamesTheZ (approved):
LGTM.

PR #264, opened 2022-04-19
Fix wrong start/end calculation in aten::slice converter

wyzero:


pytorch_blade/src/compiler/mlir/converters/impl/shapes.cpp
```diff
@@ -246,8 +246,8 @@ bool ConvertAtenSlice(
 
   auto start_ivalue = torch::jit::toIValue(jit_start);
   auto end_ivalue = torch::jit::toIValue(jit_end);
-  auto start_isnone = !start_ivalue || start_ivalue->isNone();
-  auto end_isnone = !end_ivalue || end_ivalue->isNone();
+  auto start_isnone = start_ivalue && start_ivalue->isNone();
```
fortianyou, line 249: Nit: Shall we return false `if (!start_ivalue || !end_ivalue)` ?

wyzero: I think we should not return in such case. `start_ivalue ` and `end_ivalue ` can be none if aten::slice is a dynamic slice (just like the ut I added in this PR), and we definitely want to support such case.

fortianyou (approved):
LGTM

pytorch_blade/src/compiler/mlir/converters/impl/shapes.cpp
fortianyou: After discussing with @wyzero , if `torch::jit::toIValue` does not return a true value then it means that we can't get it at conversion time, which means that it's a dynamic slice.

PR #273, opened 2022-04-20
build tao bridge with bazel

Orion34-lanbo:
Replace cmake logics from `tao_build.py` for `tao_bridge`'s build and test


.github/workflows/main.yml
```diff
@@ -120,49 +120,6 @@ jobs:
       run: |
         set -e
         bash ./scripts/ci/deploy_tensorflow_blade.sh
-  CUDA11-TORCH171:
```
Orion34-lanbo, line 123: Removed for debug now, will be restored when tests are passed

third_party/bazel/tf_protobuf/tf_protobuf_workspace.bzl.tpl
```diff
@@ -1,14 +1,18 @@
...
 
 def _tf_protobuf_workspace():
+    # NOTE(lanbo.llb): tf_http_archive will check if rule with same already exists,
+    # if so, the duplicate rule by `tf_http_archive` will not be used.
+    # Our own version of protobuf will be used instead of tf_community's.
+    # Make sure this rule is loaded first when building tf's custom op/passes.
     tf_http_archive(
-        name = "tf_protobuf",
-        patch_file = ["//third_party/bazel/tf_protobuf:protobuf.patch"],
+        name = "com_google_protobuf",
```
qiuxiafei, line 9: It seems that it's renamed to overwrite tf's `com_google_protobuf` intentionally, but I didn't get the motivation...

scripts/python/tao_build.py
```diff
@@ -222,6 +227,95 @@ def configure(root, args):
...
+    with cwd(tao_bazel_root), gcc_env(args.bridge_gcc):
+        # make sure version.h is generated
+        execute("bazel build --config=release //:version_header_genrule")
+
+    with cwd(root):
+        # copy version.h from tao_bridge
+        # NOTE(lanbo.llb): This is no longer needed when tao_compiler is build
+        # in workspace `org_tao_compiler` instead of `org_tensorflow`
+        execute(
+            f"cp {tao_bazel_root}/bazel-bin/tao_bridge/version.h tao_compiler/decoupling/version.h"
```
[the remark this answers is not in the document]
Orion34-lanbo: This is because `version.h` here is for `tao_compiler` build, which is compiled in `tf_community` dir. Later when `tao_compiler` can be built in `tao_compiler` dir, this no longer needed.

tao/WORKSPACE
```diff
@@ -0,0 +1,41 @@
...
+load("@//:workspace0.bzl", "tao_bridge_workspace0")
+
+tao_bridge_workspace0()
+
+load("@local_config_tf_protobuf//:tf_protobuf_workspace.bzl", "tf_protobuf_workspace")
+
+tf_protobuf_workspace()
+
+# Initialize TensorFlow's external dependencies.
+load("@org_tensorflow//tensorflow:workspace3.bzl", "tf_workspace3")
```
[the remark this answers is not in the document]
Orion34-lanbo: We have specifically used `com_google_protobuf` `com_google_googletest` `zlib` `cuda` `rocm` configs. But it is important to load tf's workspace for using the same `BUILD` file for both `tao_bridge` and `tao_compiler` build, since the `BUILD` will load plenty things from `org_tensorflow`. Besides, the workspace files should all be loaded for it to work.

tao_compiler/mlir/xla/ral/BUILD
```diff
@@ -532,6 +532,7 @@ cc_library(
     name = "common_context_bridge",
     srcs = [
         "context/common_context_impl.cc",
+        "//tensorflow/compiler/mlir/xla:compile_metadata_bridge_genproto",
```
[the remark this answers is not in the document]
Orion34-lanbo: I think this will still be necessary when we `have the whole DISC compiler and bridges as a workspace out of tensorflow`.

third_party/bazel/tf_protobuf/tf_protobuf_workspace.bzl.tpl
Orion34-lanbo: Since the `protobuf` we get from github urls already have `bazel` build files, it's workspace is `com_google_protobuf `. If we manually changes this name by adding a patch, it would require some extra works, so I take the easy way to use the same rule name as in `org_tensorflow`, thus we can use the bazel patch/BUILD/bzl files directly from release branch like r2.4.0-rc3/r1.15.0-rc5.

scripts/ci/build_and_test.sh
```diff
@@ -30,16 +30,16 @@ BLADE_DISC_DIR=tao/python/blade_disc_tf
   && rm -rf tao/build \
   && cd tf_community && bazel clean --expunge)
 
-python ${ENTRY} ${VENV_PATH} -s configure --bridge-gcc default --compiler-gcc default ${CPU_ONLY}
-python ${ENTRY} ${VENV_PATH} -s build_tao_bridge ${CPU_ONLY}
+python ${ENTRY} ${VENV_PATH} -s configure --bridge-gcc default --compiler-gcc default ${CPU_ONLY} --bazel_bridge
```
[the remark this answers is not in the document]
Orion34-lanbo: done

scripts/ci/build_and_test.sh
```diff
@@ -17,17 +17,14 @@ SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 source ${SCRIPT_DIR}/parse_args.sh "$@"
 
-if [[ -f ~/.cache/proxy_config ]]; then
```
qiuxiafei, line 20: Why remove it? It's done in `build_and_test.sh` instead?

scripts/python/tao_build.py
```diff
@@ -206,7 +212,6 @@ def configure(root, args):
              envs += " ACL_ROOT_DIR={} ".format(acl_root)
```
qiuxiafei, line 212: bad indent ...

scripts/python/tao_build.py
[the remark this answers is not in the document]
qiuxiafei: No better solution for now ..

tao/tao_bridge.bzl
```diff
@@ -23,6 +23,9 @@ def tao_bridge_cc_test(
             "@local_config_tf//:libtensorflow_framework",
         ],
         data = data,
-        linkstatic = linkstatic,
+        linkstatic = select({
+            "@local_config_blade_disc_helper//:is_mkldnn": 1,
```
qiuxiafei, line 27: nit: It's not appropriate to hard code a specific logic about mkldnn here in a common rule, IMO. Just pass `linkstatic=1` for mkl-related tests?

third_party/bazel/acl/acl_makefile.patch
```diff
@@ -14,4 +14,11 @@ index 0000000..bd9a183
 +	mkdir -p $(INSTALLDIR)/build $(INSTALLDIR)
 +	cp build/libarm_compute.a build/libarm_compute_core.a build/libarm_compute_graph.a $(INSTALLDIR)/build/
 +	cp -r build/src/core/arm_compute_version.embed arm_compute/ src/ utils/ support/ include/ $(INSTALLDIR)
+diff --git a/build/.gitignore b/build/.gitignore
+new file mode 100644
+index 000000000..48cdce852
+--- /dev/null
++++ b/build/.gitignore
+@@ -0,0 +1 @@
++placeholder
```
qiuxiafei, line 23: what's this for?

third_party/bazel/onednn/onednn_configure.bzl
```diff
@@ -18,9 +18,10 @@ def _onednn_impl(repository_ctx):
         repository_ctx.template("onednn.BUILD", Label("//bazel/onednn:onednn.BUILD.tpl"), {
             "%{ACL_ROOT}": "\"ACL_ROOT_DIR\": \"{}\",".format(acl_root) if with_aarch64 else "",
             "%{ACL_SETTING}": "\"DNNL_AARCH64_USE_ACL\": \"ON\"," if with_aarch64 else "",
-            "%{CXX11_SETTING}": "\"USE_CXX11_ABI\": \"ON\"" if if_cxx11_abi else "",
+            "%{CXX11_SETTING}": "\"USE_CXX11_ABI\": \"ON\"," if if_cxx11_abi else "",
```
qiuxiafei, line 21: lol, text templating is always dirty.

qiuxiafei (approved):
Just some trivial comments, the whole PR LGTM.

scripts/ci/build_and_test.sh
Orion34-lanbo: This is not necessary for build servers in `CN-HK` regions, also it will cause bazel download fail for arm server in beijing or shanghai region

third_party/bazel/acl/acl_makefile.patch
Orion34-lanbo: Just make `build` dir exists before build happens, otherwise bazel build will failed.

scripts/ci/build_and_test.sh
wyzero: We need this for AARCH64 CI since it could not use CN-HK.

scripts/ci/build_and_test.sh
```diff
@@ -17,17 +17,14 @@ SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
...
-fi
-
 ENTRY=scripts/python/tao_build.py
 VENV_PATH=/opt/venv_disc
 BLADE_DISC_DIR=tao/python/blade_disc_tf
 
 # cleanup build cache
 (rm -rf build \
   && rm -rf tao/build \
+  && cd tao && bazel clean --expunge && cd .. \
```
wyzero, line 27: Why do we need to remove this? @Yancey1989 could you help to double-check this. I think the remote cache we used is enough for caching?

tao/tao_bridge.bzl
Orion34-lanbo: Currently `tao_bridge` does not have tests actually uses `mkldnn`. This is just for the linking problems.

scripts/ci/build_and_test.sh
Orion34-lanbo: I don't quite follow, you mean `bazel clean --expunge` is not necessary？

wyzero: Sorry. I missed something. It's ok.

PR #275, opened 2022-04-21
Add more flags to control cpu codegen strategy

wyzero:


tao_compiler/mlir/disc/transforms/fusion_utils.cc
```diff
@@ -991,14 +991,22 @@ class BaseCpuFusionStrategy : public BaseFusionStrategy {
   virtual StringRef getName() override { return "BaseCpuFusionStrategy"; }
 };
 
+bool enableEagerTransposeFusion() {
+  static const char* env = getenv("DISC_CPU_ENABLE_EAGER_TRANSPOSE_FUSION");
+  if (!env) return false;
+  return std::atoi(env) != 0;
```
minminsun, line 997: In case a user sets this env as "true", this function returns false?

wyzero: Thanks, done!

minminsun (approved):
LGTM

PR #280, opened 2022-04-22
[Torch] slim header files inclusion

fortianyou:
The inclusion of TorchScript one-stop header will such as:
```
#include <torch/script.h>
```
will lead to dependencies on un-related header files that are not required.
This will increase the risk of symbol conflicts. Sometimes we will run into segfault and it's hard to reason about.

To make our source files' dependencies clear, the pull request uses as more specific headers as possible.


Yancey1989 (approved):
LGTM

PR #281, opened 2022-04-22
Remove redundant copies of cluster outputs

wyzero:


tao_compiler/mlir/xla/ral/context/base/base_context.h
```diff
@@ -53,10 +53,21 @@ class BaseOutputBufferWrapper : public OutputBufferWrapper {
...
 
+  // Returns true if this wrapper is the exclusive owner
+  bool owned() const override { return owned_; };
+
+  // mark that this wrapper exclusively owns the underlying buffer.
+  void markOwned() override { owned_ = true; }
+
+  // Release the ownership of the wrapper buffer.
+  // This requires that the buffer is owned the this wrapper.
+  void release() override { deleter_ = nullptr; }
```
bddppq, line 64: Should `release()` also set `owned` to false?

pytorch_blade/src/compiler/mlir/runtime/ral_context.cpp
```diff
@@ -197,11 +197,29 @@ torch::List<torch::Tensor> RalContext::CreateAndBindingOutputs(
...
-    torch::Tensor out_tensor = IsEmptyTensor(out_buf->shape())
-        ? torch::zeros(out_buf->shape(), option)
-        : torch::from_blob(
+    torch::Tensor out_tensor;
+    if (IsEmptyTensor(out_buf->shape())) {
+      out_tensor = torch::zeros(out_buf->shape(), option);
+    } else if (out_buf->owned()) {
+#ifdef TORCH_BLADE_BUILD_WITH_CUDA
+      auto deleter = (output_info.device == "cuda")
+          ? c10::cuda::CUDACachingAllocator::raw_delete
```
bddppq, line 206: Do we hardcode to use `CUDACachingAllocator` as the cuda allocator? I think ideally we should use PyTorch's `Allocator` interface (which has `raw_allocate` and `raw_deallocate` api).

fortianyou: What is the difference between CUDACachingAllocator and Allocator?

pytorch_blade/src/compiler/mlir/runtime/ral_context.cpp
```diff
@@ -197,11 +197,29 @@ torch::List<torch::Tensor> RalContext::CreateAndBindingOutputs(
...
+      auto deleter = c10::free_cpu;
+#endif
+      out_tensor = torch::from_blob(
+          const_cast<void*>(out_buf->data()),
+          out_buf->shape(),
+          deleter,
+          option);
+      out_buf->release();
+    } else {
+      out_tensor =
```
fortianyou, line 218: Nit: Shall we decrease the number of buffer users in this case? And make the final user reuse the underline buffer

pytorch_blade/src/compiler/mlir/runtime/ral_context.cpp
bddppq: Allocator is the interface, CUDACachingAllocator is the default gpu implementation (users can implement their own cpu/gpu allocator)

tao_compiler/mlir/xla/ral/context/base/base_context.h
wyzero: Thanks! Done.

pytorch_blade/src/compiler/mlir/runtime/ral_context.cpp
wyzero: Thanks！ Done.

bddppq (approved):
LGTM

fortianyou (approved):
Overall LGTM

PR #288, opened 2022-04-26
Expand mhlo::PowOp when the exponential is positive integer

wyzero:


tao_compiler/mlir/disc/transforms/disc_algebra_simplifier.cc
```diff
@@ -0,0 +1,150 @@
...
+//   %2 = x * %1
+//   ... ...
+//   %n = x * %n-1
+struct ExpandPowOp : public OpRewritePattern<mhlo::PowOp> {
+  using OpRewritePattern<mhlo::PowOp>::OpRewritePattern;
+
+  LogicalResult expandPowOp(mhlo::PowOp op, PatternRewriter& rewriter,
+                            int64_t exponential) const {
+    // TODO(disc): support non-positive exponential
+    if (exponential <= 0) return failure();
```
minminsun, line 46: Shall we limit the max exponential to avoid bad case?

wyzero: done!

minminsun (approved):
LGTM.

PR #296, opened 2022-04-27
Support gemm pre-packing via onednn/acl on aarch64

wyzero:
ACL NEGEMM itself supports weight pre-packing by default. Without this
PR, we create a new dnnl::matmul primitive for each matmul call, which
in turn creates a new NEGEMM object inside the primitive on aarch64.
We can not re-use the pre-packed weight from previous matmul call even
for the same matmul configuration since the NEGEMM is destroyed after
each matmul call. This RP tries to cache the matmul primitive across
matmul calls and thus can make use of the underlying pre-packed weight
for compatible matmul calls.

### DIFF
--- a/tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
+++ b/tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
@@ -114,6 +114,17 @@ bool enableWeightPrePacking() {
   return enabled;
 }
 
+int initWeightPrePackingCacheCapacity() {
+  const char* env = getenv("DISC_CPU_WEIGHT_PRE_PACKING_CACHE_CAPACITY");
+  if (!env) return 1000;
+  return std::atoi(env);
+}
+
+int getWeightPrePackingCacheCapacity() {
+  static int capacity = initWeightPrePackingCacheCapacity();
+  return capacity;
+}
+
 }  // namespace
 
 using ideep::data_type;
@@ -582,11 +593,6 @@ struct OnednnACLGemmKey {
   }
 };
 
-std::thread::id getThreadId() {
-  static thread_local std::thread::id tid = std::this_thread::get_id();
-  return tid;
-};
-
 struct OnednnACLKeyHasher {
   std::size_t operator()(const OnednnACLGemmKey& key) const {
     std::size_t seed = std::hash<int>()(key.m);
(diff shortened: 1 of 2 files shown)


minminsun:
> This RP tries to cache the matmul primitive across matmul calls and thus can make use of the underlying pre-packed weight for compatible matmul calls.

How can we reuse the pre-packed weight without the weight_is_const info?

wyzero:
> > This RP tries to cache the matmul primitive across matmul calls and thus can make use of the underlying pre-packed weight for compatible matmul calls.
> 
> How can we reuse the pre-packed weight without the weight_is_const info?

ACL NEGEMM will save packed second operand unconditionally by default.
https://github.com/ARM-software/ComputeLibrary/blob/master/arm_compute/core/Types.h#L1974

minminsun:
> 

Shall we rely on the `weight_is_const` to decide whether to reuse the cached matmul primitive?

wyzero:
> > 
> 
> Shall we rely on the `weight_is_const` to decide whether to reuse the cached matmul primitive?

I'am still WIP, thus the code you saw previously is not the final version.

wyzero:
@minminsun The code is ready for review a.t.m.

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
```diff
@@ -593,6 +648,28 @@ void onednn_ral_gemm(ExecutionContext* ctx, void* stream_handle,
...
+  MatmulPrimitive* primitive;
+  {
+    OnednnACLGemmKey key{m,    n,    k,      1,
+                         tp_a, tp_b, B.data, std::this_thread::get_id()};
+    std::lock_guard<std::mutex> l(state->mu);
+    primitive = getOrCreateMatmulPrimitive(state->cached_primitive, key, src,
+                                           weight, output);
+  }
+  ideep::matmul_forward::compute(*primitive, src, weight, output);
+#else
```
minminsun, line 677:
```suggestion
#elif defined(TAO_X86)
```

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
```diff
@@ -556,6 +557,74 @@ struct OnednnGemmState : public Context::Resource {
...
+    hash_combine(seed, key.transpose_b);
+    hash_combine(seed, key.const_weight_ptr);
+    hash_combine(seed, key.tid);
+    return seed;
+  }
+};
+
+using MatmulPrimitive = ideep::matmul_forward::super;
+using OnednnACLGemmCache =
+    std::unordered_map<OnednnACLGemmKey, std::shared_ptr<MatmulPrimitive>,
```
minminsun, line 606: Does each MatmulPrimitive hold a pre-packed weight? If yes, this cache will consume a huge amount of memory in dynamic shape cases with wide shape ranges, e.g. batch size from 1 to 1000.

wyzero: Yes, for sure. This is why such feature is not enabled by default. we eventually need a LRU cache for cached primitive.

wyzero: More challenging problem is primitive is not thread safe, leading to consuming more memory in multi-thread.

minminsun: OK, we should keep in mind this risk and be very careful in applying this feature on ARM cpus.

minminsun (approved):

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
wyzero: I found there is a lru_cache implementation in ideep. I'll change the code to use this implementation.

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
wyzero: done

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
wyzero: done

PR #298, opened 2022-04-27
Rebase community TensorFlow 20220424.

JamesTheZ:


fortianyou (approved):
LGTM

tao_compiler/mlir/disc/IR/disc_ral_ops.td
```diff
@@ -29,7 +29,6 @@ def RAL_Dialect : Dialect {
 
 class RAL_Op<string mnemonic, list<Trait> traits> :
     Op<RAL_Dialect, mnemonic, traits> {
-  let verifier = [{ return Verify(*this); }];
```
wyzero, line 32: Why do we remote this check here?

tao_compiler/mlir/xla/ral/context/init_stream_executor.cc
```diff
@@ -86,7 +86,8 @@ se::Stream* GetOrCreateDefaultCudaStreamExecutorStream(int device_ordinal,
     // StreamExecutor use cuda primary context.
     // A non-primary context was not verfied by StreamExectuor.
     // So we would not support non-primary cuda context currently.
-    e.stream = new se::Stream(executor, new CUDAStream(cuda_executor, stream));
+    e.stream = new se::Stream(
+        executor, std::make_unique<CUDAStream>(cuda_executor, stream), false);
```
wyzero, line 90: See the comment in line 85, unique-ptr may be not safe?

tao_compiler/mlir/disc/IR/disc_ral_ops.td
JamesTheZ: Because previous verify function does nothing but returning an OK directly.

tao_compiler/mlir/xla/ral/context/init_stream_executor.cc
JamesTheZ: The new se::Stream only accepts unique-ptr after rebasing the community TF. When e.stream is not destroyed, the created unique-ptr held by e.stream is safe.

PR #300, opened 2022-04-28
Support mkl gemm pre-packing on X86

wyzero:


tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
```diff
@@ -607,21 +581,117 @@ struct OnednnACLKeyHasher {
...
+                      CblasPacked, m, n, k, reinterpret_cast<Tinput*>(A.data),
+                      A.strides[0], reinterpret_cast<Tweight*>(packed_weight),
+                      B.strides[0], 0.0, reinterpret_cast<Toutput*>(C.data),
+                      C.strides[0]);
+#endif
+}
+
+struct OnednnGemmState : public Context::Resource {
+  std::mutex mu;
+  std::unordered_map<opaque_t, std::vector<ideep::tensor>> packed_weight_cache;
```
minminsun, line 680: Better to replace the `std::unordered_map<opaque_t, std::vector<ideep::tensor>>` with a LRU cache which takes `{weight_pointer, weight_desc}` as the key. This could be done in another PR.

minminsun (approved):

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
```diff
@@ -586,15 +560,15 @@ struct OnednnACLGemmKey {
   // To enable large parallelism, we cache the primitive per-thread.
   std::thread::id tid;
```
fortianyou, line 561: Will cblas gemm mutate the pre-pack weights? Otherwise, why should we worry about the thread-safe?

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
```diff
@@ -607,21 +581,117 @@ struct OnednnACLKeyHasher {
...
+  opaque_t packed_weight_ = nullptr;
+};
+
+MklPackedWeight::MklPackedWeight(ExecutionContext* ctx, const CpuGemmKey& key)
+    : ctx_(ctx->getContext()),
+      driver_(ctx->getDriver<cpu::CPUDriver>(cpu::CPUDriver::name())) {
+  size_t packed_weight_size =
+      cblas_sgemm_pack_get_size(CblasBMatrix, key.m, key.n, key.k);
+  packed_weight_ = driver_->raw_alloc(ctx_, packed_weight_size);
+  cblas_sgemm_pack(
```
fortianyou, line 607: Does that mean the pre-pack is a strategy that is related to the shapes of the input and transpose operations?

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
wyzero: For cblas gemm is thread safe (thus we always use the same thread id for all instances of cblas CpuGemmKey). This thread is filed is used for onednn acl, where the primitive is not thread safe.

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
wyzero: Yes, we need to take transpose attribute into consideration as the document of mkl shown. Here are some exampels:
https://www.intel.com/content/www/us/en/developer/articles/technical/introducing-the-new-packed-apis-for-gemm.html

PR #308, opened 2022-05-06
add DynamicConvOp in ElementTypeConverterPass

minminsun:


tao_compiler/mlir/disc/transforms/element_type_converter.cc
```diff
@@ -157,6 +157,38 @@ struct ConvertReduceOpWithSmallWidthIntType
     return success();
   }
 };
+struct ConvertDynamicConvOp : public OpRewritePattern<mhlo::DynamicConvOp> {
```
wyzero, line 160: nit: Maybe add some file-check based uts & e2e UTs as well?

tao_compiler/mlir/disc/transforms/element_type_converter.cc
```diff
@@ -157,6 +157,38 @@ struct ConvertReduceOpWithSmallWidthIntType
...
+  using OpRewritePattern<mhlo::DynamicConvOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(mhlo::DynamicConvOp op,
```
wyzero, line 163: What about mhlo::ConvOp?

tao_compiler/mlir/disc/disc_compiler.cc
```diff
@@ -277,6 +265,18 @@ LogicalResult LowerHLOToLLVM(ModuleOp m, const DISCLoweringOptions& options) {
   pm.addNestedPass<FuncOp>(
       disc_ral::createDiscElementTypeConverterPass(enable_fp16));
 
+  pm.addNestedPass<FuncOp>(disc_ral::createDiscConvRewriter());
```
wyzero, line 268: Is there any special reason for which you move the amp pass forwad?

We convert mhlo::conv to mhlo::DynamicConv in this pass as well. So maybe we should re-consider the order between this pass and this `AMP` pass.

minminsun: ConvRewriter pass only transforms fp16 conv to NHWC for gpu, so the conversion from fp32 to fp16 should happen before this pass.

wyzero: nit: maybe we could consider split the `ConvRewriter` pass into two passes, one for mhlo::conv to mhlo::dynamic conv   and one for layout optimization?

tao_compiler/mlir/disc/transforms/element_type_converter.cc
```diff
@@ -157,6 +157,38 @@ struct ConvertReduceOpWithSmallWidthIntType
     return success();
   }
 };
```
JamesTheZ, line 159: Just curious that clang-formatter allows no new line between two structs.

tao_compiler/mlir/disc/transforms/element_type_converter.cc
JamesTheZ: May use template to deal with both ConvOp and DynamicConvOp:
`
template <typename OpTy>
struct ConvertConvOp : public OpRewritePattern<OpTy> {
...;
dyn_cast ...;
}
`

minminsun: Seems like DynamicConvOp has one more operand than ConvOp.

tao_compiler/mlir/disc/disc_compiler.cc
minminsun: Is this still nesscessary if mhlo::ConvOp is also covered in AMP pass?

wyzero: I think it's ok as well. In general, it would be better if we could handle mhlo::conv and mhlo::dynamic_conv in one pass and then the other passes do not need to worry about the difference between these two ops.

tao_compiler/mlir/disc/transforms/element_type_converter.cc
```diff
@@ -158,6 +158,42 @@ struct ConvertReduceOpWithSmallWidthIntType
...
 
+template <typename OpTy>
+struct ConvertConvOp : public OpRewritePattern<OpTy> {
+  using OpRewritePattern<OpTy>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(OpTy op,
+                                PatternRewriter& rewriter) const override {
+    Location loc = op.getLoc();
+    Value lhs = op->getOperand(0);
+    Value rhs = op->getOperand(1);
```
qiuxiafei, line 169: just curious about why only convert first 2 operands ...

minminsun: ConvOp only has 2 operands. The third operand of DynamicConvOp is padding size.

tao_compiler/mlir/disc/transforms/element_type_converter.cc
minminsun: Done.

tao_compiler/mlir/disc/transforms/element_type_converter.cc
minminsun: Done.

tao_compiler/mlir/disc/transforms/element_type_converter.cc
```diff
@@ -206,6 +238,7 @@ struct ElementTypeConverterPass
     patterns.insert<ConvertReduceOpWithSmallWidthIntType>(&ctx);
     if (enable_fp16_gemm_) {
       patterns.insert<ConvertDotGeneralOp>(&ctx);
+      patterns.insert<ConvertDynamicConvOp>(&ctx);
```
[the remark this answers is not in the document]
minminsun: Done.

tao_compiler/mlir/disc/tests/regression/element_type_converter.cc
```diff
@@ -0,0 +1,51 @@
...
+#include "tensorflow/compiler/mlir/disc/tests/mlir_test.h"
+#include "tensorflow/core/platform/test.h"
+
+namespace mlir_test {
+
+const std::string c_ft_path =
+    "tensorflow/compiler/mlir/disc/tests/regression/data/";
+
+TEST(ElementTypeConverterTest, Gemm) {
+  setenv("TAO_MLIR_ENABLE_AMP", "true", 1);
```
JamesTheZ, line 26: This test actually cannot guarantee that AMP is enabled. Is there any way to check that?

JamesTheZ: Or we can also rely on `[tao_compiler/mlir/disc/transforms/tests/element_type_converter.mlir](https://github.com/alibaba/BladeDISC/pull/308/files#diff-2bcd783bd2afe5ce7cd08a2cb6061dd0eb00ccf81c5d1b78ce569f019c0679d6)` for for the checking.

wyzero: File-check based UT?

JamesTheZ: > File-check based UT?

File-check does not guarantee the cpp UT also works. Just like the checking of GPU stitch fusion. But AMP should be more robust than GPU stitch. The file-check based UT should be OK.

JamesTheZ (approved):
LGTM.

PR #318, opened 2022-05-10
Support `blade gemm` for torch-blade

Orion34-lanbo:
1. Support `blade gemm` for `torch-blade`
2. Build or download `mkldnn` `onednn` `acl` and `blade gemm` directly from bazel, we do not need to call cmake in `common_setup.py` for `torch-blade` build.


pytorch_blade/bazel_build.py
```diff
@@ -66,6 +66,10 @@ def __init__(self, *args, **kwargs):
             "--action_env TORCH_BLADE_TORCH_INSTALL_PATH={}".format(self.torch_dir),
             # Workaroud issue: https://github.com/bazelbuild/bazel/issues/10327
             "--action_env BAZEL_LINKLIBS=-lstdc++",
+            "--action_env CC={}".format(which("gcc")),
+            "--action_env CXX={}".format(which("g++")),
+            # for onednn cmake external build
+            "--action_env IF_CXX11_ABI={}".format(int(self.GLIBCXX_USE_CXX11_ABI)),
```
fortianyou, line 72: Nit: I prefer to add a config such as `--config=torch_cxx11abi` that will set both config of `cxx11abi_1` and `--action_env IF_CXX11_ABI=1` in `.bazelrc`

pytorch_blade/torch_blade_build.py
```diff
@@ -59,6 +59,18 @@ def __init__(
         self.static_tensorrt = check_env_flag(
             "TORCH_BLADE_BUILD_TENSORRT_STATIC", "OFF"
         )
+        self.blade_gemm = check_env_flag(
+            "TORCH_BLADE_BUILD_BLADE_GEMM", "OFF"
+        )
+        self.blade_gemm_nvcc = os.getenv(
+            "TORCH_BLADE_BUILD_BLADE_GEMM_NVCC", "/usr/local/cuda-11.6/bin/nvcc"
```
fortianyou, line 66: Why should this `nvcc` be specific to 11.6? I prefer to move those less general configurations from `bazel_setup.py` script to `.bazelrc`. I mean these environment variables:
`TORCH_BLADE_BUILD_BLADE_GEMM_NVCC_ARCHS`
`TORCH_BLADE_BUILD_BLADE_GEMM`
`TORCH_BLADE_BUILD_BLADE_GEMM_NVCC`
`TORCH_BLADE_BUILD_BLADE_GEMM_LIBRARY_KERNELS`

pytorch_blade/bazel_build.py
Orion34-lanbo: done

pytorch_blade/ci_build/build_pytorch_blade.sh
```diff
@@ -42,10 +42,9 @@ function ci_build() {
     if [ "$TORCH_BLADE_BUILD_WITH_CUDA_SUPPORT" = "ON"  ]; then
       export TORCH_BLADE_BUILD_TENSORRT=ON
       export TORCH_BLADE_BUILD_TENSORRT_STATIC=${TORCH_BLADE_BUILD_TENSORRT_STATIC:-OFF}
-      python3 ../scripts/python/common_setup.py
-    else
-      python3 ../scripts/python/common_setup.py --cpu_only
     fi
+    # only need symlink files in common_setup.py now
+    python3 ../scripts/python/common_setup.py
```
qiuxiafei, line 47: Good ~~

pytorch_blade/torch_blade_build.py
Orion34-lanbo: As we have disscuss offline, make these action_envs into internal .bazelrc.

Orion34-lanbo: As we have disscuss offline, make these action_envs into internal .bazelrc.

tao_compiler/mlir/xla/ral/BUILD
```diff
@@ -161,15 +162,31 @@ cc_library(
     deps = if_platform_alibaba([":real_disc_patine_client_deps"])
 )
 
+cc_library(
+    name = "blade_gemm_deps",
+    deps = if_torch_disc(
```
qiuxiafei, line 167: What's the difference between `@blade_gemm` and `//tao/blade_gemm`, I am a little bit confused...
And why use `@blade_gemm` when torch_disc is `ON` ?

tao_compiler/mlir/xla/ral/BUILD
```diff
@@ -200,6 +217,60 @@ cc_library(
     ],
 )
 
+# depend on mkldnn and onednn from bazel workspace
+cc_library(
+    name = "disc_mkl_bazel",
```
qiuxiafei, line 222: renamed from `disc_mkl_bridge` but remove static mkl libs, why ....

tao_compiler/mlir/disc/disc.bzl
```diff
@@ -62,3 +62,9 @@ def if_disc_x86(if_true, if_false=[]):
         "//tensorflow/compiler/mlir/disc:disc_x86": if_true,
         "//conditions:default": if_false
     })
+
+def if_torch_disc(if_true, if_false=[]):
+    return select({
+        "//tensorflow/compiler/mlir/disc:is_torch_disc": if_true,
```
qiuxiafei, line 68: As I guess, `is_torch_disc` should be set to `true` when building for torch and `false` when for tf?
I thinks this situation could be improved to support both torch/tf in a single bazel configuration.

tao_compiler/mlir/xla/ral/BUILD
Orion34-lanbo: `//tao/blade_gemm` is from a `BUILD` files write for cmake-prebuild blade gemm lib and headers build by python scripts, `@blade_gemm` is pure bazel target which doesn't depend on any pre-bazel cmake actions.

Orion34-lanbo: `//tao/blade_gemm` is from a `BUILD` files write for cmake-prebuild blade gemm lib and headers build by python scripts, `@blade_gemm` is pure bazel target which doesn't depend on any pre-bazel cmake actions.

tao_compiler/mlir/xla/ral/BUILD
Orion34-lanbo: Originally, we don't need to depend on entire static mkl libs, only the selected ones between `start-group` and `end-group` is enough. If we depend on static mkl libs, a `multiple definitions` problems will be caused when building for torch_blade.

Orion34-lanbo: Originally, we don't need to depend on entire static mkl libs, only the selected ones between `start-group` and `end-group` is enough. If we depend on static mkl libs, a `multiple definitions` problems will be caused when building for torch_blade.

tao_compiler/mlir/disc/disc.bzl
Orion34-lanbo: Yes, set `is_torch_disc` makes toch-blade‘s entire build actions no need to directly call `cmake`. Since `tf` side's compiler still build in `org_tensorflow` workspace, so we need `is_torch_disc ` to set them apart.

tao_compiler/mlir/xla/ral/BUILD
```diff
@@ -161,15 +162,31 @@ cc_library(
...
 )
 
+cc_library(
+    name = "blade_gemm_deps",
+    deps = if_torch_disc(
+        [
+            "@blade_gemm//:blade_gemm"
+        ],
+        [
+            "//tao/blade_gemm:blade_gemm"
```
minminsun, line 172: As offline discussed, the tf tao_compiler target has no dependency on this target. If the torch ral_context target goes to the if_true branch of if_torch_disc, maybe this if_false branch as well as the build_blade_gemm() in tao_build.py can be removed?

tao_compiler/mlir/disc/disc.bzl
qiuxiafei: But this behavior makes our build processor more complicated. Considering a developer just configure bazel build once for the whole repo with device/GCC/.. and other options and then `bazel build //tao` / `bazel build //torch_blade` / `bazel build //tf_blade` in any order. Setting an option before building torch_blade will introduce extra cognitive load to developer, while it is avoidable...

tao_compiler/mlir/xla/ral/BUILD
qiuxiafei: so would the cmake branch and the cmake building code in `common_setup.py` get removed?

tao_compiler/mlir/xla/ral/BUILD
qiuxiafei: The `multiple definitions` problem seems related to whole-archive behavior in `tf_community`.
When `@mkl_static//:mkl_static_lib` is add to `deps`, these static libs will be wrapped by `whole-archive` flags in generated build scripts. While it's removed, the actual working link options are those in `linkopts`, which as no `whole-archive` flags .

Orion34-lanbo: Maybe better ways is to make a cc_library for each static lib from `@mkl_static//:mkl_static_lib` instead of making all of them as one `cc_library`

tao_compiler/mlir/xla/ral/BUILD
Orion34-lanbo: Removed in latest commit for this pr

tao_compiler/mlir/disc/disc.bzl
Orion34-lanbo: This is no longer needed anymore

tao_compiler/mlir/xla/ral/BUILD
Orion34-lanbo: While we do not need this for `tao_compiler_main` build, `disc-replay-main` still need this, still need some refactor for `disc-replay-main` to be able to build.

Orion34-lanbo:
I have revert the commit of using bazel build onednn since I have encountered a wired core dump when calling functions from acl. This core dump will be dealt with in another pr, which will not block our monthly release.

scripts/python/common_setup.py
```diff
@@ -17,6 +17,7 @@
 import logging
 import sys
 import re
+import json
```
qiuxiafei, line 20: not needed?

tao_compiler/mlir/disc/disc.bzl
qiuxiafei: But still used at `tao_compiler/mlir/xla/ral/BUILD:250` in this PR.

scripts/python/common_setup.py
Orion34-lanbo: used for `deduce_cuda_version`

scripts/python/tao_build.py
```diff
@@ -452,7 +452,8 @@ def bazel_build(target, flag=""):
 
         bazel_build(TARGET_TAO_COMPILER_MAIN, flag=flag)
         bazel_build(TARGET_DISC_OPT, flag=flag)
-        bazel_build(TARGET_DISC_REPLAY, flag=flag)
+        if not (args.platform_alibaba and args.blade_gemm):
```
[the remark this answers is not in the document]
Orion34-lanbo: `disc-replay-main` will depend on `ral`, it is compiled in `org_tensorflow` making it impossible to use `@blade_gemm`. It is not need for internal build now, so I make it like this.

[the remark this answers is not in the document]
Orion34-lanbo: This condition is removed now.

tao_compiler/mlir/xla/ral/BUILD
qiuxiafei: `if_torch_disc` is better to be eliminated, it's better if there's nothing special for any of torch or tf.

tao_compiler/mlir/xla/ral/BUILD
qiuxiafei: I'm trying to remove tf_communit's whole archive.

qiuxiafei (approved):

PR #321, opened 2022-05-11
remove whole archive in tf_community

qiuxiafei:


pytorch_blade/src/compiler/mlir/converters/BUILD
```diff
@@ -43,10 +42,11 @@ cc_test(
     ],
     linkstatic = True,
     deps = [
-        "//src:libtorch_blade.so",
-        "@zlib",
+        ":torch_blade_mhlo_converter",
```
fortianyou, line 45: Why change to static link?

pytorch_blade/src/BUILD
```diff
@@ -106,6 +106,7 @@ pybind_library(
     ] + if_tensorrt_enabled([
         "@tensorrt//:nvinfer_headers",
     ]),
+    alwayslink = 1,
```
fortianyou, line 109: Is the alwayslink must be set to true.

third_party/bazel/tensorrt/repo.bzl
```diff
@@ -11,7 +11,7 @@ cc_import(
 
 cc_import(
     name = "myelin_executor_static",
-    static_library = "libmyelin_executor_static_patched.a",  # use patched one
+    static_library = "lib/libmyelin_executor_static.a",  # use patched one
```
fortianyou, line 14: Nit: remove the `use patched one`

fortianyou (commented):
How is the `--whole-archive` removed from `tf_community`?

pytorch_blade/src/BUILD
qiuxiafei: Yes, since this pybind library is not directly referenced by any C++ code, it will be dropped when linking if the default whole-archive behavior is disabled.

third_party/bazel/tensorrt/repo.bzl
qiuxiafei: done

pytorch_blade/src/compiler/mlir/converters/BUILD
qiuxiafei: As discussed offline, this test should depend on //src:libtorch_blade.so to test agains the finally output shared library. Reverted.

fortianyou (approved):
LGTM

PR #334, opened 2022-05-24
[NFC] refactor disc supported op list in TF bridge side

wyzero:


minminsun (approved):
LGTM, thanks

tao/tao_bridge/passes/tao_mark_for_compilation_pass.cc
```diff
@@ -2556,6 +2556,114 @@ std::unordered_set<string>* GetFusedOpTable() {
...
+
+#ifndef TF_1_12
+  ops.insert(ops.end(), {
+    "AddV2",
+    "BatchMatMulV2",
+    "GatherV2"
+  });
+#endif
+  // clang-format on
+  return ops;
```
qiuxiafei, line 2664: `ops` could still be a `static` local right? Ops of 1.12 could still be guarded by the macro.

wyzero: The return value of this function is stored into a static variable. Thus we do not need to use `static` here.

PR #341, opened 2022-05-26
Support tf.QuantizeV2 and tf.Dequantize

wyzero:


tao_compiler/mlir/disc/disc_compiler.cc
```diff
@@ -753,34 +757,106 @@ LogicalResult LowerHLOToSharedLibrary(ModuleOp m,
...
 
-  mlir::SmallVector<std::unique_ptr<mlir::Pass>, 1> lower_passes;
-  lower_passes.emplace_back(mlir::disc_ral::createDiscLowerTfPass());
-  CreateConvertMlirToXlaHloPipeline(
-      tf2xla, "XLA_CPU_JIT",
-      /*prefer_tf2xla=*/false,
-      /*custom_legalization_passes=*/lower_passes);
+  // Note that the region-based control-flow produced here still contains
+  // function call ops which get inlined by the subsequent inliner pass.
+  pm.addPass(mlir::TF::CreateTFFunctionalControlFlowToRegions());
```
minminsun, line 781: Are these added passes related to the support of quantize/dequantize ops?

wyzero: Before this PR, we directly call `CreateConvertMlirToXlaHloPipeline ` to get a pre-defined set of passes to do tf2mhlo conversion. The pre-defined pass pipeline is not able to support customize quantized op lowering, thus we have to build our own pass pipeline.

minminsun: Seems like it's the InlinerPass in the CreateConvertMlirToXlaHloPipeline that breaks the lowering of (de)quantize op?

wyzero: Actually it's `CreateLegalizeTfTypesPass`. This pass assume that all quantized type can be safely removed, which means we have to rewriter all quantized ops before this pass. However, the pre-defined pass pipeline does not support register customized quantized op lowering patterns.

minminsun (approved):
LGTM

PR #347, opened 2022-05-31
Release GPU memory after UT to prevent OOM, enhance error code message.

JamesTheZ:
After adding a set of variables for UTs, it results in OOM on some GPUs due to the lack of memory freeing. This PR frees the unused GPU memory for the UTs.


Yancey1989 (approved):
LGTM

PR #364, opened 2022-06-08
[Torch] remove corrupt passes

fortianyou:


pytorch_blade/torch_blade/pass_manager.py
```diff
@@ -358,6 +358,3 @@ def _jit_pass_hack_cpu_device(graph):
         if prim_const.hasAttribute('value') and prim_const.kindOf('value') == 's' and prim_const.s('value') == 'cpu':
             prim_const.s_('value', 'cuda')
 
-
-OptPipelines.register_pipeline("LICM", _jit_pass_licm)
```
qiuxiafei, line 362: Still keep this: https://github.com/alibaba/BladeDISC/blob/41f02b6149b84fea17272f6595e97b740d9f097e/pytorch_blade/src/compiler/jit/licm.cpp#L97 ?

fortianyou: YES, the pass might be used in another place, but the optimization pipeline related to the pass is deleted.

qiuxiafei (approved):
LGTM

PR #367, opened 2022-06-08
[Torch] fix EngineClass global initialization

fortianyou:


pytorch_blade/src/compiler/backends/engine_class.cpp
```diff
@@ -160,82 +160,88 @@ c10::intrusive_ptr<EngineClass> EngineClass::Deserialize(
...
-                  -> c10::intrusive_ptr<EngineClass> {
-                // A convenient way to instantiate an object and get an
-                // intrusive_ptr to it is via `make_intrusive`. We use
-                // that here to allocate an instance of
-                // MyStackClass<std::string> and call the single-argument
-                // std::vector<std::string> constructor with the serialized
-                // state.
-                return EngineClass::Deserialize(std::move(serialized));
-              });
+  static std::once_flag flag;
```
qiuxiafei, line 163: Does this function still have to be exposed to pybind if it called by global initializer...

fortianyou: Already removed.

pytorch_blade/src/pybind.cpp
```diff
@@ -49,7 +49,6 @@ std::string pybind_version() {
 
 template <>
 void initModules<false>(py::module& m) {
-  torch::blade::backends::InitTorchBladeEngine();
```
qiuxiafei, line 52: How about `InitBladeDiscEngine`? There seems no global initializer for it.

fortianyou: This will be used in TorchDISC & TorchBlade

Orion34-lanbo (approved):

pytorch_blade/src/pybind.cpp
qiuxiafei: Does a C++ user of TorchBlade still have to call `InitBladeDiscEngine` before `torch::jit::load` a optimized model ?

fortianyou: Nop

PR #380, opened 2022-06-09
GPU kLoop fusion fully vectorization optimization.

JamesTheZ:
1. vectorizes both load and store.
2. vectorizes some fp16 computation instructions.
3. interleaves computation instructions that are not vectorized.


tao_compiler/mlir/disc/transforms/codegen_utils.cc
```diff
@@ -44,17 +47,27 @@ int getRowReductionScheduleHint(Operation* op) {
 
 int getVectorizeOrTileHint(Operation* op) {
   assert(op);
-  lmhlo::FusionOp fusion = op->getParentOfType<lmhlo::FusionOp>();
-  if (!fusion) return 1;
+  lmhlo::FusionOp fusion = dyn_cast<lmhlo::FusionOp>(op);
+  if (!fusion) {
```
qiuxiafei, line 51: when would `fusion` be null?

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
```diff
@@ -446,8 +446,9 @@ struct DiscSpecializeFusionWithSpeculationPass
     Location loc = fusion_op.getLoc();
 
     Value out_element_number;
```
qiuxiafei, line 448: Could be removed?

tao_compiler/mlir/disc/transforms/lhlo_legalize_roots_to_loops.cc
```diff
@@ -326,6 +314,51 @@ LogicalResult lowerWithScheduleLoop(
...
+  const auto loc = dominant_op->getLoc();
+  OpBuilder b(root_ops.back());
+  Value zero = b.create<arith::ConstantIndexOp>(loc, 0);
+  Value one = b.create<arith::ConstantIndexOp>(loc, 1);
+  Value vec_size = b.create<arith::ConstantIndexOp>(loc, vector_size);
+  auto num_elements = emitNumElementsComputation(b, loc, dominant_op);
+  auto thread_number = b.create<arith::DivUIOp>(loc, num_elements, vec_size);
+  Value var;
+  SmallVector<Value, 2> vars;
+  scf::ParallelOp simt_loop = createParallelAndSetInsPt(
```
qiuxiafei, line 330: newbie question: is it here the `ParallelOp` where the vectorization magic happens?

tao_compiler/mlir/disc/transforms/codegen_utils.cc
JamesTheZ: If `op` is not a `lmhlo::FusionOp`, the cast returns null.

tao_compiler/mlir/disc/transforms/lhlo_legalize_roots_to_loops.cc
JamesTheZ: Not the ParallelOp. The ParallelOp is to form the SIMT parallelism structure (i.e. grid and block) of GPU kernel.
What helps to do the vectorization is the ForOp at line-336 bellow. After unroll and interleaving for this op, the instructions inside this ForOp could be vectorized after some LLVM passes.

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
JamesTheZ: Done. Thanks.

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
```diff
@@ -461,48 +461,66 @@ struct DiscSpecializeFusionWithSpeculationPass
...
-      threshold = b.create<arith::ConstantIndexOp>(loc, max_threads_per_wave);
       FusionPatternBase fusion_pattern(fusion_op);
       Operation* dominant_equivalent_op = fusion_pattern.getRootOps().back();
-      out_element_number =
+      Value out_element_number =
           emitNumElementsComputation(b, loc, dominant_equivalent_op);
+      if (mem_intensive_opt_experimental) {
+        vector_size = 4;
+      }
+      Value divisible = b.create<arith::CmpIOp>(
```
minminsun, line 487: Should we also check whether the buffer address divisible by vector_size?

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
```diff
@@ -461,48 +461,66 @@ struct DiscSpecializeFusionWithSpeculationPass
...
+      pred = b.create<arith::AndIOp>(loc, larger, divisible);
     } else if (fusion_type == FusionType::kLoop) {
-      threshold = b.create<arith::ConstantIndexOp>(loc, max_threads_per_wave);
       FusionPatternBase fusion_pattern(fusion_op);
       Operation* dominant_equivalent_op = fusion_pattern.getRootOps().back();
-      out_element_number =
+      Value out_element_number =
           emitNumElementsComputation(b, loc, dominant_equivalent_op);
+      if (mem_intensive_opt_experimental) {
+        vector_size = 4;
```
minminsun, line 485: Maybe the vector_size should depend on the element type, e.g. vector_size = 128 / element_bit_width ?

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
minminsun: I see AssumeAlignmentOp inserted to annotate the buffer with the given alignment, so this check is not that critical. But since this annotation is only best-effort, it's still better to do the address alignment check if it doesn't hurt the performance.

JamesTheZ: Very good suggestion. Checking all buffers of a fusion could be expensive, instead, I will add alignment attribute of `memref.alloc` to ensure the alignment of allocated buffers.

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
JamesTheZ: The vector-size also affects the thread level parallelism. I will check the performance of different sizes when vector-size is 8 for FP16.

qiuxiafei:
Another question, vectorization increases pressure on register usage, will it bring regression in some case?

tao_compiler/mlir/disc/transforms/lhlo_legalize_roots_to_loops.cc
qiuxiafei: Thanks.

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
minminsun: OK, this is a tradeoff. I think it's OK to leave the vector-size decision for fp16 as a TODO for now.
AFAIK there's not any benifit to have a vector width larger than 128 bit, e.g. vector_szie=4 for fp64. What do you think?

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
minminsun: That's a good idea. But I didn't see any aligned alloc implemention in RAL. I don't know if I missed something... @wyzero

JamesTheZ:
> Another question, vectorization increases pressure on register usage, will it bring regression in some case?

No worry for register pressure of kLoop fusion. Even given a high vector-size, it will use a very small set of registers due to the short lift time of each register.

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
JamesTheZ: Updated. Current commit set the vector-size according to 128 / min-bit-width-of-outputs. I have check the performance. It is good for a small shape like `128 x 1024`. The case of tiny shape hurts (e.g., `8 x 1024`), but the penalty is less than a us, which I think is OK.

minminsun: Great!

tao_compiler/mlir/disc/transforms/disc_specialize_fusion_with_speculation.cc
wyzero: I think it’s better to use a uniform alignment  setting (e.g. a large enough setting) across all allocations if there are not so many small buffers. Mixed alignment settings may introduce redundant copies, which in turn may lead to performance drop. 

For TensorFlow users,  A default alignment value 64 is used according to following link.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/allocator.h#L114

For PyTorch users,  A default alignment value 16 or 64 is used according to following link.
https://github.com/pytorch/pytorch/blob/master/c10/core/alignment.h

For gpu users, it seems that `cudaMalloc` uses a alignment 256 by default according to following link.
https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses

Thus, I think currently we do not need to explicitly set alignment.

minminsun (approved):
LGTM

PR #388, opened 2022-06-14
bugfix: skip non-shape-tensor from_elements op in shape opt pass

wyzero:


tao_compiler/mlir/disc/transforms/tests/disc-shape-optimization.mlir
```diff
@@ -224,3 +224,18 @@ func @main(%arg0: tensor<?x?x?xf32>) -> tensor<?x?xf32> {
...
+// regression test: non-shape-tensor from_element op
+
+// CHECK-LABEL: @main
+// CHECK-SAME: (%[[ARG0:.*]]: i32) -> (tensor<i32>, tensor<2x1xi32>
+func @main(%arg0 : i32) -> (tensor<i32>, tensor<2x1xi32>) {
+  // CHECK: %[[T0:.*]] = tensor.from_elements %[[ARG0]] : tensor<i32>
+  // CHECK: %[[T1:.*]] = tensor.from_elements %[[ARG0]], %[[ARG0]] : tensor<2x1xi32>
+  // CHECK: return %[[T0]], %[[T1]]
+  %0 = tensor.from_elements %arg0 : tensor<i32>
+  %1 = tensor.from_elements %arg0, %arg0 : tensor<2x1xi32>
```
minminsun, line 238: I'm not familiar with `from_elements` op. According to `isCandidateShapeTensorType()`, seems like `%0` is not shape tensor because its rank is 0, while `%1` is not shape tensor because its rank is 2. If that's true, I'm curious in which case `from_elements` will generate shape tensor with rank=1?

wyzero: `from_elements` can be used to generate a tensor which is not a shape tensor. Actually, it can generate tensor with any rank and any element type according to its definition.

```
func @main(%arg0 : f32, %arg1 : tensor<f32>) -> tensor<f32> {
  %0 = tensor.from_elements %arg0 :  tensor<f32>
  %1 = mhlo.mul %0, %1 ...
  return %1 ....
}

```

minminsun (approved):
LGTM

PR #389, opened 2022-06-14
support quantized_conv2 with const zero bias with qint32 type

wyzero:


tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -660,10 +661,32 @@ Type ToLegalElementType(Type type) {
...
+  return true;
+}
+
 bool isConstantZeroTensor(Value v) {
   auto castOp = dyn_cast_or_null<TF::CastOp>(v.getDefiningOp());
   if (castOp) return isConstantZeroTensor(castOp->getOperand(0));
 
+  auto ty = v.getType().dyn_cast<RankedTensorType>();
+  if (ty && ty.getElementType().isa<mlir::TF::Qint32Type>()) {
+    return isQint32ConstantZeroTensor(v);
```
minminsun, line 687: Is the num_elements = 1 or num_elements > 1 allowded if it's qint32 type?

wyzero: I add a check to protect empty tensor. `num_elements` > 1 is ok.

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -660,10 +661,33 @@ Type ToLegalElementType(Type type) {
...
+  return (numElems != 0);
+}
+
 bool isConstantZeroTensor(Value v) {
   auto castOp = dyn_cast_or_null<TF::CastOp>(v.getDefiningOp());
   if (castOp) return isConstantZeroTensor(castOp->getOperand(0));
 
+  auto ty = v.getType().dyn_cast<RankedTensorType>();
+  if (ty && ty.getElementType().isa<mlir::TF::Qint32Type>()) {
+    return isQint32ConstantZeroTensor(v);
```
qiuxiafei, line 688: why Qint32 is allowed to to have multiple value in tensor?

wyzero: The `i32` or `float` case also supports zero tensors with `num elements > 1`. Here zero tensors are those contents are all zeros, no matter how many elements it has.  

The different between the `f32/i32` case and `qint32` case is:
- The content of a qint32 tensor is a tensorflow tensor object, which does not support `splat` storage.
- The content of a i32/f32 tensor is a mlir dense attributes, which supports `splat` storage. If all the elements of a tensor are zeros, then it could use splat storage, in which case we only need to check the first elements.

PR #398, opened 2022-06-17
support custom install requires for torch blade

chenbohua3:


pytorch_blade/setup.py
```diff
@@ -125,6 +125,12 @@ def run(self):
 
 
 install_requires = ["networkx", "onnx>=1.6.0", f"torch=={torch.__version__}"]
+custom_install_requires = os.getenv("TORCH_BLADE_CUSTOM_INSTALL_REQUIRES", None)
```
bddppq, line 128: why do we need this option?

chenbohua3: torch_blade packages for some internal users need to be introduced with some additional dependencies. This option make it possible to customize these dependencies when package torch_blade.

PR #400, opened 2022-06-17
thread safe acl conv and depthwise conv

wyzero:
Most part of this PR is just copied from the corresponding ACL layer
implementations. we re-implement these layers to add thread safety
support.


Yancey1989 (approved):
LGTM, better to add a unit test for multi-thrads.

qiuxiafei (approved):
LGTM

PR #431, opened 2022-06-29
simplify patterns: `x*1=x` and `x+0=x`

wyzero:


tao_compiler/mlir/disc/transforms/disc_algebra_simplifier.cc
```diff
@@ -30,6 +31,63 @@ namespace mlir {
...
+  }
+  DenseElementsAttr denseAttr;
+  if (!matchPattern(tensor, m_Constant(&denseAttr))) return false;
+  if (denseAttr.getNumElements() != 1 && !denseAttr.isSplat()) return false;
+
+  Type elemTy = denseAttr.getElementType();
+  if (elemTy.isIntOrIndex()) {
+    return (*denseAttr.getValues<APInt>().begin()).getSExtValue() == elemVal;
+  } else if (elemTy.isa<FloatType>()) {
+    return (*denseAttr.getValues<APFloat>().begin()).convertToDouble() ==
```
minminsun, line 51: nit: why convert to double for comparison with an int64_t value?

minminsun:
How much perf gain observed with these simplifications?

minminsun (approved):
LGTM

tao_compiler/mlir/disc/transforms/disc_algebra_simplifier.cc
wyzero: 1, convert f16/f32/f64 to a unified type f64. 
2, We only support the case where `elemVal` is int64_t (as indicated by its type), thus it does not matter: c++ implicit conversion will cast to f64.
3, If we need to support `elemVal`  to double, we need to overload this function.

fortianyou, line 51: It's weird to compare the equality of double and int.

tao_compiler/mlir/disc/transforms/disc_algebra_simplifier.cc
```diff
@@ -30,6 +31,63 @@ namespace mlir {
...
+           elemVal;
+  }
+  return false;
+}
+
+// `x + 0` or `0 + x` could be simplified to `x`
+struct AddZeroTensorOp : public OpRewritePattern<mhlo::AddOp> {
```
fortianyou, line 58: How about adding patterns for `x - 0` and something similar.

wyzero:
> How much perf gain observed with these simplifications?

These patterns are used to remove some redundant `biasadd` (e.g. bias is zero) or pattern like `alpha * x * y + beta, where alpha = 1, beta = 0`. The performance improvement from the these patterns may not significant. However it can help us to simplify the graph, which in turn enables more opportunities.

tao_compiler/mlir/disc/transforms/disc_algebra_simplifier.cc
wyzero: It may be added in the future.  However, some similar patterns may not be safe. For example, `x * 0 = 0` is not true if `x` is `NaN`.

tao_compiler/mlir/disc/transforms/disc_algebra_simplifier.cc
qiuxiafei: A template is simple enough? Also feels weird to compare double and int.

PR #461, opened 2022-07-13
[CI] fix install python3 on ubuntu20

fortianyou:
Before the deploy docker building failed because of the installation of python3 on ubuntu20. And we want to use the daily built runtime docker images for the TorchBlade benchmark mentioned in #450


qiuxiafei (approved):
LGTM

PR #476, opened 2022-07-18
add opset option for onnx2trt conversion

Orion34-lanbo:
For `tensorflow==1.15`, onnx2trt conversion will encounter `Attribute not found: axes` errors with the default `opset` for `tf2onnx==1.11.1` and `onnx==1.11.0`, which is 13. We should set a default and tested opset value for onnx2trt conversion

### DIFF
--- a/tensorflow_blade/tf_blade/gpu/tf_to_trt.py
+++ b/tensorflow_blade/tf_blade/gpu/tf_to_trt.py
@@ -105,6 +105,7 @@ class Tf2TrtOpt:
         new_input_names: List[str],
         subgraph_outputs: List[str],
         dynamic_shapes: List[Shape],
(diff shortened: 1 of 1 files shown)


tensorflow_blade/tf_blade/gpu/tf_to_trt.py
```diff
@@ -105,6 +105,7 @@ def _convert_subgraph_to_onnx(
         new_input_names: List[str],
         subgraph_outputs: List[str],
         dynamic_shapes: List[Shape],
+        opset: int,
```
qiuxiafei, line 108: nit: `opset_version` ?

qiuxiafei (approved):

PR #483, opened 2022-07-20
[TorchMhlo] Add fp16 support

fortianyou:
Torch-MLIR hasn't supported fp16 conversion before. The pull request adds the fp16 support.


Yancey1989 (approved):
LGTM++

PR #488, opened 2022-07-21
support compilation cache serialization in disc TF plugin

wyzero:
Support load/store compilation cache on disk in disc TF plugin. Note that:
- we suppose the clustering result is stable across different process.
The above assumption may not always hold. Thus be careful when using this
feature. 

- We may have multiple instances of `TaoCompilationCache` if we have
multiple sessions in the same process. However there should be only one
instance of `PersistentCompliationCache` in order to get a stable result.

### DIFF
--- a/tao/tao_bridge/common.cc
+++ b/tao/tao_bridge/common.cc
@@ -45,8 +45,8 @@ static void AllocateTaoBridgeFlags() {
                               &opts->verbose_compilation_err_log));
   CHECK_OK(ReadBoolFromEnvVar("TAO_ENFORCE_VERBOSE_COMPILATION_LOG", false,
                               &opts->verbose_compilation_log));
-  CHECK_OK(ReadStringFromEnvVar("TAO_CACHE_DUMP_PATH", "",
-                                &opts->tao_cache_dump_path));
+  CHECK_OK(ReadStringFromEnvVar("DISC_COMPILATION_CACHE_DUMP_PATH", "",
+                                &opts->disc_cache_dump_path));
   CHECK_OK(ReadBoolFromEnvVar("TAO_ENABLE_MLIR", false, &opts->enable_mlir));
 
   CHECK_OK(ReadBoolFromEnvVar("TAO_MLIR_BRANCH_ONLY", false,
--- a/tao/tao_bridge/common.cc
+++ b/tao/tao_bridge/common.cc
@@ -45,8 +45,8 @@ static void AllocateTaoBridgeFlags() {
                               &opts->verbose_compilation_err_log));
   CHECK_OK(ReadBoolFromEnvVar("TAO_ENFORCE_VERBOSE_COMPILATION_LOG", false,
                               &opts->verbose_compilation_log));
-  CHECK_OK(ReadStringFromEnvVar("DISC_COMPILATION_CACHE_DUMP_PATH", "",
-                                &opts->disc_cache_dump_path));
+  CHECK_OK(ReadStringFromEnvVar("DISC_COMPILATION_CACHE_PATH", "",
+                                &opts->disc_cache_path));
   CHECK_OK(ReadBoolFromEnvVar("TAO_ENABLE_MLIR", false, &opts->enable_mlir));
 
   CHECK_OK(ReadBoolFromEnvVar("TAO_MLIR_BRANCH_ONLY", false,
--- a/tao/tao_bridge/kernels/tao_compilation_cache.cc
+++ b/tao/tao_bridge/kernels/tao_compilation_cache.cc
@@ -408,7 +408,7 @@ class PersistentCompliationCache {
     do {
       path =
           absl::StrCat(cache_dump_path_, "/", "disc_cache_item_", nextIdx_++);
-    } while (tensorflow::Env::Default()->FileExists(path));
+    } while (tensorflow::Env::Default()->FileExists(path).ok());
     return path;
   }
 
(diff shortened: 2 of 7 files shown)


tao/tao_bridge/common.cc
```diff
@@ -45,8 +45,8 @@ static void AllocateTaoBridgeFlags() {
                               &opts->verbose_compilation_err_log));
   CHECK_OK(ReadBoolFromEnvVar("TAO_ENFORCE_VERBOSE_COMPILATION_LOG", false,
                               &opts->verbose_compilation_log));
-  CHECK_OK(ReadStringFromEnvVar("TAO_CACHE_DUMP_PATH", "",
-                                &opts->tao_cache_dump_path));
+  CHECK_OK(ReadStringFromEnvVar("DISC_COMPILATION_CACHE_DUMP_PATH", "",
+                                &opts->disc_cache_dump_path));
```
fwd4, line 49: why some remains TAO_* and some becomes DISC_*, what is the standard here?

tao/tao_bridge/kernels/tao_compilation_cache.cc
```diff
@@ -316,6 +316,221 @@ Status GetDeviceId(OpKernelContext* ctx, int* device_id) {
...
+  static PersistentCompliationCache& Global() {
+    static PersistentCompliationCache cache(
+        GetTaoBridgeOptions()->disc_cache_dump_path);
+    return cache;
+  }
+
+  // Saves cache table to file located at `getCacheTableFilePath`.
+  Status DumpToFile() {
+    mutex_lock lock(mu_);
+    return DumpToFileLocked();
```
fwd4, line 363: maybe use the name DumpToFileUnsafe?

tao/tao_bridge/kernels/tao_compilation_cache.cc
```diff
@@ -316,6 +316,221 @@ Status GetDeviceId(OpKernelContext* ctx, int* device_id) {
...
+  Status DumpToFileLocked();
+
+ private:
+  mutex mu_;
+  string cache_dump_path_;
+  using DeviceLevelCache =
+      std::unordered_map<ItemSignature, string, ItemSignature::Hash>;
+  // target device str -> device level cache.
+  std::unordered_map<std::string, DeviceLevelCache> cache_;
+  std::atomic<uint64_t> nextIdx_;
```
qiuxiafei, line 426: not camel-cased name ?

tao/tao_bridge/common.cc
wyzero: For the new envs, better to use DISC_*.  For the old envs, we gradually rename it to DISC_* on demand？

tao/tao_bridge/kernels/tao_compilation_cache.cc
```diff
@@ -1156,33 +1248,44 @@ Status TaoCompilationCache::CompileImpl(
...
-        PrepareCompilerInput(function, constant_args, fixed_shape_args,
-                             host_args, variable_args, ctx, &input, is_mlir);
-    TF_RETURN_IF_ERROR(entry->compilation_status);
 
+    bool hit_cache = false;
     std::string output_file_name;
-    if (!tensorflow::Env::Default()->LocalTempFilename(&output_file_name)) {
-      return errors::Internal(
-          "couldn't get temp tao_compiler_result file name");
+    if (!disc_cache_dump_path_.empty()) {
```
qiuxiafei, line 1254: nit: if `disc_cache_dump_path_` is also used for loading, how about just call it `disc_cache_path_` ?

tao/tao_bridge/kernels/tao_compilation_cache.cc
```diff
@@ -316,6 +316,221 @@ Status GetDeviceId(OpKernelContext* ctx, int* device_id) {
...
+  // Returns the name of file used to store `CompilationCacheResult`.
+  static string getCacheTableFileName() { return "disc_cache"; }
+
+  // Returns the path of file used to store `CompilationCacheResult`.
+  string getCacheTableFilePath() {
+    return absl::StrCat(cache_dump_path_, "/", getCacheTableFileName());
+  }
+
+  // Returns the singleton
+  static PersistentCompliationCache& Global() {
```
qiuxiafei, line 354: How about removing copy ctor of `PersistentCompliationCache` to prevent miss using like:
```
PersistentCompliationCache cache = PersistentCompliationCache::Global()
```

tao/tao_bridge/kernels/tao_compilation_cache.cc
wyzero: just try to follow TF community practice.
https://github.com/tensorflow/tensorflow/search?q=Locked&type=code

image

qiuxiafei (approved):
Generally LGTM with little trivial comments~

tao/tao_bridge/kernels/tao_compilation_cache.cc
wyzero: done, thanks.

tao/tao_bridge/kernels/tao_compilation_cache.cc
wyzero: done

tao/tao_bridge/kernels/tao_compilation_cache.cc
wyzero: done

qiuxiafei (approved):
LGTM

PR #489, opened 2022-07-21
impl for supporting f.Bucketize op at compiler side

Orion34-lanbo:


tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -1080,6 +1083,161 @@ LogicalResult convertQuantizedConv2DWithBiasAndRequantizeOp(func::FuncOp func) {
   return success();
 }
 
+// Returns 1D 64-bit dense elements attribute with the given values.
+static mlir::DenseIntElementsAttr GetI64ElementsAttr(
```
qiuxiafei, line 1087: nit: do we have a common place for this kind of utility.

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -1080,6 +1083,161 @@ LogicalResult convertQuantizedConv2DWithBiasAndRequantizeOp(func::FuncOp func) {
...
+    absl::Span<const int64_t> values, Builder* builder) {
+  auto ty = mlir::RankedTensorType::get({static_cast<int64_t>(values.size())},
+                                        builder->getIntegerType(64));
+  return mlir::DenseIntElementsAttr::get(
+      ty, llvm::makeArrayRef(values.data(), values.size()));
```
qiuxiafei, line 1092: Seems not safe to make a ref to `values` array? Callers may a pass temporary one.

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -1080,6 +1083,161 @@ LogicalResult convertQuantizedConv2DWithBiasAndRequantizeOp(func::FuncOp func) {
...
+}
+
+// Gets all float values from the given attribute and push them to `values`.
+static DenseFPElementsAttr GetF32ElementsAttr(Attribute attr, Builder *builder) {
+  SmallVector<float, 4> values;
+  auto array_attr = attr.cast<ArrayAttr>();
```
qiuxiafei, line 1098: have a check on `array_attr`?

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -1080,6 +1083,161 @@ LogicalResult convertQuantizedConv2DWithBiasAndRequantizeOp(func::FuncOp func) {
...
+    // Bucketize kernel. However, comparing an int32 or int64 against float may
+    // lead to inaccurate bucketing due to rounding.
+    if (input_type.isF64()) {
+      boundaries = rewriter.create<mhlo::ConvertOp>(loc, boundaries, rewriter.getF64Type());
+    } else {
+      input = rewriter.create<mhlo::ConvertOp>(loc, input, rewriter.getF32Type());
+    }
+
+    // get input type again since it maybe converted
+    input_type = input.getType().dyn_cast<RankedTensorType>();
```
qiuxiafei, line 1172: nit: just  move this line to `else` above?

tao_compiler/mlir/disc/tests/tensorflow_ops/bucketize.cc
```diff
@@ -0,0 +1,93 @@
...
+
+namespace mlir_test {
+
+const std::string c_ft_path =
+    "tensorflow/compiler/mlir/disc/tests/tensorflow_ops/data/";
+
+// TODO(lanbo.llb): Support more input data types
+
+// f32 static shape test case
+TEST(TFBucketizeOpTest, StaticShapeF32) {
```
qiuxiafei, line 28: nice tests~

Orion34-lanbo:
> 

For `1` tf's upstream doesn't have one for `BucketizeOp` now.
As for `2`, this is currently only the `compiler` side code, we still need to support clustering for `BucketizeOp` in `tao_bridge`, maybe put the toggle there is better?

tao_compiler/mlir/disc/transforms/lower_tf.cc
Orion34-lanbo: I think maybe in some utils files is better, but I don't know if we already have one

wyzero: disc_util.h|cc?

tao_compiler/mlir/disc/tests/tensorflow_ops/bucketize.cc
```diff
@@ -0,0 +1,93 @@
...
+  EXPECT_TRUE(feature_test_main(
+      /*mlir_file_path*/ c_ft_path + "bucketize_s_f32.mlir",
+      /*backend_types*/ { BackendType::kX86 },
```
wyzero, line 31: For BuketizeOp  itself,  it should be ok on ARM backend or cuda backend as well?  maybe we can add these backends as well.

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -1080,6 +1083,161 @@ LogicalResult convertQuantizedConv2DWithBiasAndRequantizeOp(func::FuncOp func) {
   return success();
 }
 
+// Returns 1D 64-bit dense elements attribute with the given values.
+static mlir::DenseIntElementsAttr GetI64ElementsAttr(
+    absl::Span<const int64_t> values, Builder* builder) {
```
wyzero, line 1088: nit: better to use llvm::ArrayRef<int64_t>

tao_compiler/mlir/disc/tests/tensorflow_ops/bucketize.cc
Orion34-lanbo: Currently `BuketizeOp` support cuda backend, should be added. For arm, I haven't done some tests on arm, but the op should support arm as well.

tao_compiler/mlir/disc/transforms/lower_tf.cc
Orion34-lanbo: According to llvm [docs](https://mlir.llvm.org/doxygen/classmlir_1_1Attribute.html#a711073f32e43429d5faaada8c387ff88), `cast` API has a `assert` check first. Thus we switch to use `dyn_cast` which will return a nullptr when type is not compatible.

tao_compiler/mlir/disc/transforms/lower_tf.cc
Orion34-lanbo: done

tao_compiler/mlir/disc/transforms/lower_tf.cc
Orion34-lanbo: removed this func, use a pre-existed function

tao_compiler/mlir/disc/transforms/lower_tf.cc
Orion34-lanbo: done

tao_compiler/mlir/disc/tests/tensorflow_ops/bucketize.cc
Orion34-lanbo: modified to kSupportedBackend

tao_compiler/mlir/disc/transforms/lower_tf.cc
Orion34-lanbo: removed, use a existed function with `llvm::ArrayRef<int64_t>` input

qiuxiafei (approved):

PR #490, opened 2022-07-22
[fix] torch_ltc_disc_backend no longer exists.

qiuxiafei:


pytorch_blade/bazel_build.py
```diff
@@ -79,8 +79,6 @@ def __init__(self, *args, **kwargs):
             "--action_env TORCH_BLADE_TORCH_INSTALL_PATH={}".format(self.torch_dir),
         ]
 
-        if self.torch_major_version >= 1 and self.torch_minor_version >= 12:
```
fortianyou, line 82: Why is this PR raised? I think the target is stills used in TorchDISC.

https://github.com/alibaba/BladeDISC/blob/82c1d2a1233949a924fc2f61ab75438921d91480/pytorch_blade/.bazelrc#L17

qiuxiafei: Ooooops ...  This config exists in external code but not in internal `.bazelrc` ... The latter should be fixed.

PR #493, opened 2022-07-25
Loop invariant code motion for operators with side effect.

JamesTheZ:


tao_compiler/mlir/disc/transforms/disc_side_effect_loop_invariant_code_motion.cc
```diff
@@ -0,0 +1,119 @@
...
+
+    auto valueKnownNotModifiedInLoop = [&](Value value) {
+      return llvm::all_of(value.getUsers(), [&](Operation* user) {
+        // Check if `user` is outside of the loop.
+        if (loop.getLoopBody().findAncestorOpInRegion(*user) == nullptr) {
+          return true;
+        }
+
+        // Skip if there are memref alias operators.
+        if (disc_ral::IsMemRefAliasOp(user)) {
```
qiuxiafei, line 54: nit: to my understanding, it's still possible to analyses alias of users recursively to move more op outside?

qiuxiafei (approved):

tao_compiler/mlir/disc/transforms/disc_side_effect_loop_invariant_code_motion.cc
JamesTheZ: Yes, you are right! I do not do that because most alias ops are already moved out by the non-side-effect-LICM pass and I do not observe alias ops left for softmax and layernorm operators. But surely there are potentially some more opportunities by analyzing alias chains.

PR #496, opened 2022-07-25
bugfix: using rewriter to erase an op inside a pattern rewriter

wyzero:


tao_compiler/mlir/disc/transforms/input_inline_fusion_pattern.cc
```diff
@@ -175,7 +175,7 @@ LogicalResult InputInlineFusionPattern::processParallelOp(
     // Clean all the ops that do not have LoadOps inside the nested
     // ParallelOps and is not the ancestor of any ops that have LoadOps
     // inside the nested ParallelOps.
-    cleanUnusedLhloOps(parent_block);
+    cleanUnusedLhloOps(parent_block, &rewriter);
```
fortianyou, line 178: Are there other usages of cleanUnusedLhloOps elsewhere?

fortianyou: image

There is another cleanUnusedLhloOps definition will it introduce problems too?

fortianyou (approved):
LGTM

qiuxiafei (approved):

tao_compiler/mlir/disc/transforms/input_inline_fusion_pattern.cc
wyzero: > Are there other usages of cleanUnusedLhloOps elsewhere?

These are not inside a rewrite pattern.

wyzero: > image
> 
> There is another cleanUnusedLhloOps definition will it introduce problems too?

seems that it is a duplicate implementation. I'll remove it in the next PR.

wyzero: I double-check the code: it's inside tf_community, not inside disc. I think it's the code we upstream to the community in the last year.

Issue #514, opened 2022-08-02
[Quantization] refactor ptq of trt backend


chenbohua3:
The overall process can be divided into the following steps:

- [x] make each subgraph executable, this allows us to collect the inputs of each subgraph when inference with the Calibration data.
- [x] modify the graph and add data-collector node.
- [x] save the inputs and use trt calibrator to build the quantization engine.
- [x] drop the q_val in codebase since it is no longer needed. (QAT model comes from the blade_compression and PTQ is done by trt itself) 
- [x] remove the added node in c_module during the data collecting process.

chenbohua3:
Close this issue since all things are done:)

PR #516, opened 2022-08-02
[Quantization] Add data collector for trt ptq

chenbohua3:
This pr contains two parts mentioned #514 
+ make each subgraph executable
+ save data for each fusion group

### DIFF
--- a/pytorch_blade/torch_blade/quantization/prepare_data.py
+++ b/pytorch_blade/torch_blade/quantization/prepare_data.py
@@ -16,7 +16,7 @@ from typing import List
 import torch
 import torch_blade
 from torch_blade import utils
-from torch_blade.clustering.support_group_conversion import group_nodes
+from torch_blade.clustering.support_group_conversion import group_nodes, replace_group_with_engine
 from torch_blade.config import Config
 from torch_blade.logging import logger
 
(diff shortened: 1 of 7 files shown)


pytorch_blade/torch_blade/quantization/prepare_data.py
```diff
@@ -0,0 +1,247 @@
...
+            self.graph.appendNode(list_unpack)
+            list_unpack.moveAfter(call_method)
+            list_unpack.addInput(call_method.output())
+            list_unpack.eraseOutput(0)
+            for out in g_node.output_list():
+                lu_out = list_unpack.addOutput()
+                lu_out.setType(out.type())
+                out.replaceAllUsesWith(lu_out)
+
+            g_node.destroy()
```
fortianyou, line 132: Nit: can we reuse [torch_blade/clustering/support_group_conversion.py#L41](https://github.com/alibaba/BladeDISC/blob/main/pytorch_blade/torch_blade/clustering/support_group_conversion.py#L41)

pytorch_blade/torch_blade/quantization/prepare_data.py
```diff
@@ -0,0 +1,247 @@
...
+            list_pack.output().setType(torch_blade.tools.get_list_tensor_type())
+
+            # insert the observer, three steps:
+            #   1. register the observer
+            #   2. create getattr node for the observer
+            #   3. create call method node for the observer
+            collect_observer = DataCollectObserver()
+            collect_observer = torch.jit.script(collect_observer)
+            observer_name = f"{self.collect_observer_prefix}_{idx}"
+            all_collect_observer_name.append(observer_name)
```
fortianyou, line 180: So did here, we can reuse the similar codes written previously. Please refactor the code and move it to the right commonplace if it's needed.

pytorch_blade/torch_blade/quantization/prepare_data.py
```diff
@@ -0,0 +1,247 @@
...
+            self.c_module.inference(*data, self.custom_module_owner)
+
+        for observer_name in self.all_collect_observer_name:
+            data = self.custom_module_owner.__getattr__(observer_name).data
+            # the first one is dummy data
+            all_data.append(data[1:])
+        return all_data
+
+    def save_data(self, data):
+        tmp_dir = tempfile.mkdtemp(prefix="torch_blade")
```
fortianyou, line 224: Will this tmp file be removed after the DataPreparer is released?

fortianyou: It's better to remove the temp file after it's used.

pytorch_blade/torch_blade/quantization/prepare_data.py
chenbohua3: There might be some slight difference:
1.  The compiled engine takes List of Tensor as inputs while here we don't group the inputs into one list since the input of each `prim::FusionGroup` is not a List. 
2.  The `prim::GetAttr` should be returned.
3.  The custom module owner is the a new-added input while the `_replace_group_with_engine` uses the first original input.

Difference 2 and 3 is easy to solve by extending `_replace_group_with_engine`. However I can not found an elegant way to solve the difference 1. Any suggestions?

pytorch_blade/torch_blade/quantization/prepare_data.py
chenbohua3: Yes, while the trt ptq is enabled, codes about deleting these temp files will be added.

pytorch_blade/torch_blade/quantization/prepare_data.py
chenbohua3: done, I extended the `replace_group_with_engine` and reuse it in ` make_fusion_group_executable`

pytorch_blade/torch_blade/quantization/prepare_data.py
chenbohua3: done, I reuse the codes about inserting the observer

fortianyou (approved):
LGTM

PR #530, opened 2022-08-08
[AArch64] make ACL target arch configurable

wyzero:
1, add a new build option `target_cpu_arch` to make ACL target arch
configurable.

2, add a new ACL patch for YITIAN to enable BF16/int8 dot instruction

3, currently only works in TF bridge. Torch bridge will be supported in
the future.


qiuxiafei (approved):
LGTM

PR #536, opened 2022-08-09
[AArch64] change the default arch to armv8.6-a-sve2

wyzero:
Note that it's safe on old arm processor since acl will also do runtime
cpu feature check and dispatch correct kernel correspondingly.


wyzero:
Change the default arch is not safe for the part of code outside of ACL. Thus we plan to close this PR.

PR #547, opened 2022-08-11
tanyo/rebase tf 11

fortianyou:
TensorFlow main changes:
1. Add assembly form for `mhlo.dynamic_reshape`, `mhlo.reshape`, `mlhlo.clamp`, `mhlo.real_dynamic_slice`
2. The output type before conversion and after must be the same
3. Some minor ut fix.

LLVM main changes:
1. math::AbsOp -> math::AbsFOp

TorchMLIR changes:
+ Upstream add TorchToMhlo passes; we disable them for now. And will be enable later.

### DIFF
--- a/mhlo_builder/csrc/mlir/mhlo/builder/standard.cpp
+++ b/mhlo_builder/csrc/mlir/mhlo/builder/standard.cpp
@@ -177,7 +177,7 @@ llvm::Optional<mlir::Value> BuildCastStdConstScalarToHloConstTensor(
     return llvm::None;
   }
   const mlir::Attribute& val_attr = def.getValue();
-  auto scalar_ty = mlir::RankedTensorType::get({}, val_attr.getType());
+  auto scalar_ty = mlir::RankedTensorType::get({}, std_scalar.getType());
   auto const_attr = mlir::DenseElementsAttr::get(scalar_ty, val_attr);
   auto result = builder.create<mlir::mhlo::ConstantOp>(loc, const_attr);
   return result.getResult();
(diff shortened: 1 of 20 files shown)


mhlo_builder/csrc/mlir/mhlo/builder/standard.cpp
```diff
@@ -177,7 +177,7 @@ llvm::Optional<mlir::Value> BuildCastStdConstScalarToHloConstTensor(
     return llvm::None;
   }
   const mlir::Attribute& val_attr = def.getValue();
-  auto scalar_ty = mlir::RankedTensorType::get({}, val_attr.getType());
+  auto scalar_ty = mlir::RankedTensorType::get({}, std_scalar.getType());
```
qiuxiafei, line 180: any difference here?

pytorch_blade/src/torch-mlir/lib/Conversion/TorchToMhlo/TorchToMhlo.cpp
```diff
@@ -746,7 +757,9 @@ LogicalResult ConvertAtenOp<AtenSigmoidOp>::matchAndRewrite(
   Value negVal = rewriter.create<mhlo::NegOp>(loc, input);
   Value expVal = rewriter.create<mhlo::ExpOp>(loc, negVal);
   Value addVal = rewriter.create<mhlo::AddOp>(loc, expVal, one);
-  rewriter.replaceOpWithNewOp<mhlo::DivOp>(op, one, addVal);
+  Value output = rewriter.create<mhlo::DivOp>(loc, one, addVal);
+  rewriter.replaceOpWithNewOp<mhlo::ConvertOp>(
```
qiuxiafei, line 761: what's the intention for such `ConvertOp` added at the end?

mhlo_builder/csrc/mlir/mhlo/builder/standard.cpp
fortianyou: llvm is updated, the old api is not valid here

pytorch_blade/src/torch-mlir/lib/Conversion/TorchToMhlo/TorchToMhlo.cpp
fortianyou: Because we are dynamic shapes, some static shapes are not carried during conversion. But the latest llvm mhlo will check the output type before and after the conversion. We introduce ConvertOp to avoid the error.

qiuxiafei (approved):

PR #556, opened 2022-08-15
disable some static checks of mhlo.dot_general

wyzero:
Recently mhlo community introduces more static checks for mhlo
dot_general op. Without these change, following IR is valid:

```
mhlo.dot_general(%0, %1) : (tensor<?x?x?xf32>, <4x?x?x?f32>)
```

After the above change, it's invalid due to mismatch static batch
dimension size.

However, `tf.BatchMatmul(tensor<?x?x?xf32>, tensor<4x?x?xf32>)` is
still valid and the `tf.BatchMatmul` tf2mhlo converter does not
handle shape propagation between the lhs & rhs, leading to the failure
of the above static check.

This CL just disables the check in tf_community as a workaround.


qiuxiafei:
> This CL just disables the check in tf_community as a workaround.

Or shall we just relax the check a little bit instead of disabling them.

wyzero:
> 

Actually, our [shape optimization pass](https://github.com/alibaba/BladeDISC/blob/main/tao_compiler/mlir/disc/disc_compiler.cc#L252) will check this invariant (after global propagation) thus I think it's ok just to disable these local static checks inside DISC.

qiuxiafei (approved):

minminsun:
> This CL just disables the check in tf_community as a workaround.

What would be a "non-workaround" solution for this problem, convice the tf community to abandon this new check or  handle shape propagation in tf2mhlo converter?

fortianyou:
@wyzero I have picked your modifies to tf_community. Please wait for my PR https://github.com/alibaba/BladeDISC/pull/547. And don't rush to merge the code since it will result in a merging conflict of mine.

wyzero:
> @wyzero I have picked your modifies to tf_community. Please wait for my PR #547. And don't rush to merge the code since it will result in a merging conflict of mine.

I see #547 is used for rebase purpose, maybe we should not merge a bugfix pr with it?

wyzero:
> > This CL just disables the check in tf_community as a workaround.
> 
> What would be a "non-workaround" solution for this problem, convice the tf community to abandon this new check or handle shape propagation in tf2mhlo converter?

Two options:

- TF and  mhlo both reject `tf.BatchMatmul(tensor<?x?x?xf32>, tensor<4x?x?xf32>)` and let the frontend user generate correct IR.
- TF can accept `tf.BatchMatmul(tensor<?x?x?xf32>, tensor<4x?x?xf32>)`, rewrite most of the tf2mhlo converter to insert reshape whenever possible.

PR #571, opened 2022-08-18
fix batchmatmul params for bladnn::gemm()

minminsun:


fortianyou (commented):
Nit: How about add a UT to check it?

qiuxiafei:
> Nit: How about add a UT to check it?

These lines are built under `PLATFORM_ALIBABA` macro, maybe the tests could only be added to internal repositoryl, i think ...

minminsun:
I think it's OK to add in this repo some MatMul and Batchmatmul tests, which will run in both external ci (without bladnn) and internal ci(with bladnn). Thanks @fortianyou @qiuxiafei

qiuxiafei (approved):
LGTM~

Issue #509, opened 2022-07-29
[RFC] Rebase Torch-MLIR Upstream TorchToMhlo Pipeline


fortianyou:
Recently, we cooperated with developers from ByteDance AML on upstreaming the TorchToMhlo pass pipeline to Torch-MLIR; see RFC https://github.com/llvm/torch-mlir/issues/999.

## The Upstream Status
+ Cmake build + code structure https://github.com/llvm/torch-mlir/pull/1083
+ Torch to MHLO lowering pipeline https://github.com/llvm/torch-mlir/pull/1083
+ Operator coverage
    + Basic Ops (~35 ops)
           https://github.com/llvm/torch-mlir/pull/1092
           https://github.com/llvm/torch-mlir/pull/1116
           https://github.com/llvm/torch-mlir/pull/1123
           https://github.com/llvm/torch-mlir/pull/1127
           https://github.com/llvm/torch-mlir/pull/1134
    + View like Ops
           https://github.com/llvm/torch-mlir/pull/1090
           https://github.com/llvm/torch-mlir/pull/1099
    + Slice like Ops  https://github.com/llvm/torch-mlir/pull/1091
    + Gather Ops https://github.com/llvm/torch-mlir/pull/1104
    + Linear Ops
          https://github.com/llvm/torch-mlir/pull/1132
          https://github.com/llvm/torch-mlir/pull/1152
    + Reduction https://github.com/llvm/torch-mlir/pull/1133
    + Others
          https://github.com/llvm/torch-mlir/pull/1128
          https://github.com/llvm/torch-mlir/pull/1141

We would eventually like to support ResNet and BERT examples in our upstream schedule. 

## Schedule in BladeDISC Repo
Now I think it's ready to rebase upstream Torch-MLIR. The issue is created to track the progress.

fortianyou:
The goals that upstream to Torch-MLIR are complete. And we had rebased our TorchToMhlo pass to the one in Torch-MLIR.

PR #580, opened 2022-08-24
Rebase Torch-MLIR

fortianyou:
Related to https://github.com/alibaba/BladeDISC/issues/509

The PR updates Torch-MLIR submodule and refines the torch-to-mhlo pipeline:
1. Add DiscConvertToMhloPass that will do op conversion patterns that haven't been upstreamed.
2. Change ConvertToMhloPass to the upstream implementation and remove the old one.
3. Update file checks and unit tests in the current repo.
4. Add shape analysis on the input TorchScript graph before torch-to-mhlo conversion
5. Fix and add shape functions for aten ops.


pytorch_blade/WORKSPACE
```diff
@@ -123,6 +123,12 @@ new_local_repository(
     path = "../../platform_alibaba/blade_gemm"
 )
 
+new_local_repository(
+  name = "mlir-hlo",
+  build_file = "third_party/torch-mlir/externals/mlir-hlo/BUILD",
```
qiuxiafei, line 128: Wow, it's expected to work right?

pytorch_blade/src/torch-mlir/lib/Conversion/TorchToMhlo/DiscTorchToMhlo.cpp
```diff
@@ -0,0 +1,565 @@
+// Copyright 2022 The BladeDISC Authors. All rights reserved.
```
qiuxiafei, line 1: Is rewrite patterns in this file newly created by us or copied from community repo?

pytorch_blade/src/torch-mlir/lib/Conversion/TorchToMhlo/DiscTorchToMhlo.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+  Value leakyActivationVal = rewriter.create<mhlo::MulOp>(
+      loc, getTypeConverter()->convertType(op.getType()), input, scaleVal);
+
+  Value compareGtZero = rewriter.create<mhlo::CompareOp>(
+      loc, input, zeroVal, mhlo::ComparisonDirection::GT);
+
+  Value output = rewriter.create<mhlo::SelectOp>(
+      loc, inputTy, compareGtZero, input, leakyActivationVal);
+  rewriter.replaceOpWithNewOp<mhlo::ConvertOp>(
```
qiuxiafei, line 471: I see `ConvertOp` is intensively used, why it's always added at the end?

pytorch_blade/WORKSPACE
fortianyou: Yes

pytorch_blade/src/torch-mlir/lib/Conversion/TorchToMhlo/DiscTorchToMhlo.cpp
fortianyou: The output tensor dtype and rank cannot change after the op conversion. But this might be true in a dynamic shape scenario because it relies on shape inference, but our shape inference is not complete for now.

It's introduced to run out of type checks before and after op conversion since `mhlo::ConvertOp` tolerates different input and output types. I think we should not have the `ConvertOp`.

fortianyou: I had reduce the `ConvertOp` usages in this PR

qiuxiafei (approved):

qiuxiafei (approved):

Yancey1989 (commented):
This PR removed some patterns, such as `aten::max_in_dim`, `aten::uniform`, better to add an issue to re-added them to the codebase ASAP.

PR #588, opened 2022-08-25
support tf32 in torch_blade trt backend

zzpmiracle:


pytorch_blade/torch_blade/config.py
```diff
@@ -177,6 +177,7 @@ def __init__(self):
         self._customize_op_black_list = []
         self._customize_jit_passes = []
         self._opt_pipeline = 'DISC'
+        self._enable_tf32 = False
```
minminsun, line 180: Shall we enable tf32 by default to align with Pytorch (before 1.12)?

fortianyou: +1

pytorch_blade/torch_blade/config.py
```diff
@@ -425,6 +426,20 @@ def customize_onnx_opset_version(self, version):
...
+        :type: bool
+        :default: False
+        """
+        return self._enable_tf32
+
+    @enable_tf32.setter
+    def enable_tf32(self, val):
+        assert isinstance(val, bool), "enable_tf32 should be bool, got {}".format(type(val))
+        self._enable_tf32 = val
+
```
fortianyou, line 442: I think we can enable TF32 in TensorRT by default. And there is no need to add a new flag like `enable_tf32`. In my opinion, it's confused with too many configurations.

pytorch_blade/torch_blade/tensorrt/tensorrt_optimization.py
```diff
@@ -34,6 +34,9 @@ def _try_build_trt_engine(dyn_proto, state, dynamic_settings, q_val, *args, **kw
     elif cfg.enable_fp16:
         with flags.builder_flags_context(1 << int(flags.BuilderFlag.FP16)):
             return _trt.cvt_onnx_to_tensorrt(dyn_proto, state, dynamic_settings, q_val)
+    elif cfg.enable_tf32:
+        with flags.builder_flags_context(1 << int(flags.BuilderFlag.TF32)):
+            return _trt.cvt_onnx_to_tensorrt(dyn_proto, state, dynamic_settings, q_val)
```
fortianyou, line 39: Also, TF32 can be set union with FP16 & FP32

fortianyou (dismissed):
LGTM

pytorch_blade/torch_blade/config.py
zzpmiracle: Done.

qiuxiafei: But at least we should have a toggle to disable it when precision problem occurs? I saw it's hard-coded.

fortianyou (approved):
LGTM

PR #591, opened 2022-08-29
[fusion] add ternary element wise op list

wyzero:
Some ternary ops (e.g. select op) support a restricted implicit broadcast
semantic, that is: if one input has rank zero, then it will be
automatically broadcasted if necessary. Thus we can know that
there is no broadcast if all operands have the same rank, which means
it's also a pure element wise op.

### DIFF
--- a/tao_compiler/mlir/disc/transforms/fusion_utils.cc
+++ b/tao_compiler/mlir/disc/transforms/fusion_utils.cc
@@ -327,10 +327,28 @@ bool isElementWiseBinary(Operation* op) {
   // clang-format on
 }
 
+// Returns true if the op is an elementwise ternary lmhlo op.
+bool isElementWiseTernary(Operation* op) {
+  // Some ternary lmhlo ops (e.g. select op) suppport a restricted implicit
+  // broadcast semantic, that is: if one input has rank zero, then it will be
+  // automatically broadcasted if necessary. Thus we can know that there is no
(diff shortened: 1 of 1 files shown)


tao_compiler/mlir/disc/transforms/fusion_utils.cc
```diff
@@ -327,10 +327,28 @@ bool isElementWiseBinary(Operation* op) {
...
 
+// Returns true if the op is an elementwise ternary lmhlo op.
+bool isElementWiseTernary(Operation* op) {
+  // Some ternary lmhlo ops (e.g. select op) suppport a restricted implicit
+  // broadcast semantic, that is: if one input has rank zero, then it will be
+  // automatically broadcasted if necessary. Thus we can know that there is no
+  // broadcast if all operands have the same rank.
+  if (isa<lmhlo::SelectOp, lmhlo::ClampOp>(op)) {
+    auto ref = op->getOperand(0).getType().cast<MemRefType>();
+    for (Value v : op->getOperands().drop_front())
```
qiuxiafei, line 338: nit: `drop_front` defaults to slice the first one element, so why not just use `begin` or `[0]` to get it directly?

wyzero: index-style-access is not always supported by the iterator.

qiuxiafei (approved):

PR #592, opened 2022-08-29
refine fake quant test.

qiuxiafei:


fortianyou (commented):
Nit: Could be better to add some overflow tests

qiuxiafei:
> Nit: Could be better to add some overflow tests

This Op is just used to carry quant info and use `int32` for underlying representation, won't be overflow I think...

fortianyou (approved):

PR #594, opened 2022-08-29
improve the performance of const lookup in RAL

wyzero:
1, assign a unique index for each unique const value.

2, using the index to do fast lool up whenever possible at runtime.

### DIFF
--- a/tao_compiler/mlir/disc/transforms/disc_convert_const_to_ral.cc
+++ b/tao_compiler/mlir/disc/transforms/disc_convert_const_to_ral.cc
@@ -122,7 +122,7 @@ class DiscConstToRALPass : public DiscConstToRALPassBase<DiscConstToRALPass> {
   int num_processing_const_ops_ = 0;
   // Map each unique name of const to a unique index. Such indices are used to
   // reduce the overhead of const lookup at runtime.
-  std::unordered_map<std::string, int> name2Idx_;
+  std::unordered_map<std::string, int> name_idx_map_;
 };
 
 // llvm.mlir.global internal constant @unique_name("unique_name\00")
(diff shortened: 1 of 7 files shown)


tao_compiler/mlir/xla/ral/context/common_context_impl.h
```diff
@@ -59,19 +67,55 @@ struct RalGlobalConstantState : public tao::ral::Context::Resource {
...
+  // fast path: using a unique const index to do look up.
+  // Note that the const index is assigned to each const at compile time.
+  // The index is unique within the compiled module level.
+  //
+  // The large index suppored to use the fast path.
+  static constexpr const int kMaxNumItemsForFastPath = 8196;
+  template <typename T>
+  using ItemArray = std::array<T, kMaxNumItemsForFastPath>;
+  using ItemLookupFastPathStorage = ItemArray<Item>;
+  using ItemLookupFastPathTable = ItemArray<std::atomic<Item*>>;
```
qiuxiafei, line 84: Why need yet another array for pointer storage?

wyzero: This is used to avoid using explicit lock in the fast path. The item pointer is updated by one thread and read by other threads. To avoid read-after-write hazard, I use an atomic pointer here.

tao_compiler/mlir/disc/transforms/disc_convert_const_to_ral.cc
```diff
@@ -120,6 +120,9 @@ class DiscConstToRALPass : public DiscConstToRALPassBase<DiscConstToRALPass> {
   LogicalResult convertConstantOp(ConstantOp const_op, MetadataProto* proto);
 
   int num_processing_const_ops_ = 0;
+  // Map each unique name of const to a unique index. Such indices are used to
+  // reduce the overhead of const lookup at runtime.
+  std::unordered_map<std::string, int> name2Idx_;
```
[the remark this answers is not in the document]
wyzero: thanks, done.

qiuxiafei (approved):

PR #595, opened 2022-08-29
[AArch64] support ACL fast math (BF16/F32 mixed-precision) for fp32 conv

wyzero:


tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
```diff
@@ -176,6 +176,20 @@ struct MkldnnConvState : public Context::Resource {
 };
 
 #if defined(TAO_AARCH64)
+
+// Returns true is ACL AMP is enabled (`DISC_CPU_ACL_USE_AMP`)
+bool useAclAMP() {
+  static bool flag = []() {
+    const char* env = getenv("DISC_CPU_ACL_USE_AMP");
```
fortianyou, line 183: Nit: tensorflow::ReadBoolFromEnvVar

wyzero: We do not want to introduce the TF dependency in RAL whenever possible.

fortianyou: How about implement something like the function utility in RAL?

wyzero: OK, I personally prefer to do this in a separate PR, including changing other places of using env flags. I'll send the PR later.

fortianyou (approved):
LGTM

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
fortianyou: Thanks!

PR #600, opened 2022-08-30
remove calibration method to avoid core dump

chenbohua3:


pytorch_blade/torch_blade/quantization/prepare_data.py
```diff
@@ -186,14 +186,19 @@ def insert_observer(self):
 
     def get_calib_data_for_each_group(self):
         all_data = []
-        self.c_module.create_method_from_graph("inference", self.graph)
+        self.c_module.create_method_from_graph("_inference", self.graph)
```
fortianyou, line 189: What is the root cause of the core dump when using the name `inference`?

fortianyou (approved):
LGTM

PR #603, opened 2022-09-01
thread local api_func cache to reduce the chance of lock conflict

wyzero:
assign a unique ID to each thread, add use a thread local api_func map
to reduce the chance of lock conflict.

### DIFF
--- a/tao_compiler/mlir/xla/ral/ral_context.cc
+++ b/tao_compiler/mlir/xla/ral/ral_context.cc
@@ -17,6 +17,8 @@
 
 #include "tensorflow/compiler/mlir/xla/ral/ral_context.h"
 
+#include <array>
+#include <atomic>
 #include <iostream>
 #include <thread>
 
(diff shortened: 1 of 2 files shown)


tao_compiler/mlir/xla/ral/ral_context.cc
```diff
@@ -27,12 +29,23 @@
...
+}
+
 const char* kRalRecvInput = "ral_recv_input";
 const char* kRalSendOutput = "ral_send_output";
 const char* kRalCudaConst = "ral_const";
 const char* kRalHostConst = "ral_const";
 const char* kRalBitcast = "inc_ref";
 
+int ThreadLocalIndex::Get() {
+  static thread_local int index = nextThreadIdx++;
```
fortianyou, line 45: Why not use `std::thread::id` directly? https://en.cppreference.com/w/cpp/thread/thread/id

wyzero: `std::thread::id` is an opaque type and thus can not be used to index an array directly. Without the above design, we have to build a hash_map to record the relationship between `std::thread::id` and the array index, which introduces more overhead.

fortianyou (approved):
Thanks, LGTM

tao_compiler/mlir/xla/ral/ral_context.cc
```diff
@@ -27,12 +29,23 @@
 namespace tao {
 namespace ral {
 
+namespace {
+
+std::atomic<int> nextThreadIdx{0};
```
qiuxiafei, line 34: nit: better to encapsulate it into `ThreadLocalIndex` as static member?

Issue #567, opened 2022-08-18
Initial quantization support in BladeDISC


wyzero:
We're now planing to start support quantization in BladeDISC. The basic idea is:
- introduce a fake_quant op. The op will be used to passing quantization related info (e.g. scale/zero_point) from frontend to BladeDISC.
- The input of BladeDISC is still a FP32 graph, with some fake_quant ops to annotate which set of ops need to be quantized.
-  It's the responsibility of BladeDISC to replace the fake_quant op and the corresponding FP32 ops (e.g. Gemm/Conv) to the quantized versions. 

The task list is shown as follow:
[] TensorFlow/PyTorch fake_quant custom ops and their corresponding implementation in TF/PyTorch Dialect.
[] mhlo_disc fake_quantize/quantize/de-quantize IR definition.
[] mhlo_disc QCONV/QMATMUL IR defintion.
[] TF/Torch custom fake-quant -> disc fake quant.
[] Fake_quantize + Conv + Fake_quantize  -> QConv pattern.
[] Fake_quantize + GEMM + Fake_quantize  -> QGEMM pattern.
[] Fake_quantize lowering.
[] Quantize/de-quantize op lowering patterns.
[] Placement rule for Qconv/Qmatmul.
[] (low priority) QConv/QMatmul layout optimization.
[] QConv/QMatmul shape inference implementation.
[] QConv/QMatmul lowering to library call implementation.
[] QConv/QMatmul kernel.

PR #608, opened 2022-09-05
[quantization] lower mhlo_disc.quantize op

wyzero:
to #567

### DIFF
--- a/tao_compiler/mlir/disc/transforms/disc_lower_quantize_and_dequantize.cc
+++ b/tao_compiler/mlir/disc/transforms/disc_lower_quantize_and_dequantize.cc
@@ -1,0 +1,124 @@
+// Copyright 2022 The BladeDISC Authors. All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/StringRef.h"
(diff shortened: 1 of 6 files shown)


tao_compiler/mlir/disc/transforms/disc_lower_quantize_and_dequantize.cc
```diff
@@ -0,0 +1,124 @@
...
+                              zeroPointTy.getEncoding());
+    Value castedZeroPoint = rewriter.create<mhlo::ConvertOp>(
+        loc, castedZeroPointTy, op.zero_point());
+    Value bcastedZeroPoint = rewriter.create<mhlo::DynamicBroadcastInDimOp>(
+        loc, inputTy, castedZeroPoint, inputShape, op.axis());
+    auto quantMinMaxTy = RankedTensorType::get({}, inputTy.getElementType());
+    Value quantMin = rewriter.create<mhlo::ConstantOp>(
+        loc, DenseElementsAttr::get(
+                 quantMinMaxTy,
+                 static_cast<float>(static_cast<int64_t>(op.quant_min()))));
```
deeply, line 66: why casting quant_min to int64_t and then float, instead of casting to float directly?

wyzero: Just in case quant_min/max is a negative integer. uint64_t -> int64_t -> float.

deeply: I see, it's safer to cast these params to int64_t firstly.
A further question is, should we guarantee quant_min/max to be stored and interpreted as int64_t? In IR definition? Seems no need to use uint64_t for these params anyway.

wyzero: it's [signless](https://discourse.llvm.org/t/what-are-signless-integers/2165) int attribute in IR definition actually, while it'll be interpreted as uint64_t by default.

deeply: I see, we are using signless int by default...

deeply (approved):
LGTM

PR #609, opened 2022-09-06
[quantization] convert tf.DiscFakeQuant to  mhlo_disc.fake_quant

qiuxiafei:
sub task of #567


tao/tao_bridge/passes/tao_mark_for_compilation_pass.cc
```diff
@@ -2576,6 +2576,7 @@ std::vector<string> GetDiscSupportedOps() {
     "Conv2D",
     "DepthwiseConv2dNative",
     "DynamicStitch",
+    "DiscFakeQuant",
```
wyzero, line 2579: nit: better to use alphabetical order?

tao/tao_bridge/ops/fake_quant_op.cc
```diff
@@ -58,7 +47,6 @@ class DiscFakeQuantOp : public OpKernel {
   bool signed_;
```
wyzero, line 47: nit: maybe better to used the same name as the attribute name?

tao_compiler/mlir/disc/IR/disc_tf_additional_ops.cc
```diff
@@ -0,0 +1,73 @@
...
+    return emitOpError("quant_min must not be less than ")
+           << expect_min << " under " << num_bits()
+           << " bits and signed=" << use_signed() << ", but got: " << q_min;
+  }
+  if (q_max > expect_max) {
+    return emitOpError("quant_max must not be greater than ")
+           << expect_max << " under " << num_bits()
+           << " bits and signed=" << use_signed() << ", but got: " << q_max;
+  }
+  if (use_dynamic()) {
```
wyzero, line 40: The op definition support `dynamic quantization`. we may choose to not add a check here for the ` use_dynamic` filed. An alternative solution is to add these checks in the corresponding rewriter pattern (e.g. when lowering quantize/de-quantize ops). It seems weird we add a filed while we can not set the attribute actually.

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -52,6 +53,13 @@ namespace mlir {
 namespace disc_ral {
 namespace {
 
+mlir::DenseIntElementsAttr getI64ElementsAttr(ArrayRef<int64_t> values,
```
wyzero, line 56: use the version in `disc_util.h`?

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -643,6 +651,30 @@ LogicalResult ConvertDequantizeOp::dequantize(TF::DequantizeOp op,
...
+    auto loc = op.getLoc();
+    auto input = op.input();
+    auto scale = op.scale();
+    auto zero_point = op.zero_point();
+    SmallVector<int64_t> axis;
+    for (auto& v : op.axis()) {
+      axis.push_back(v.cast<IntegerAttr>().getInt());
+    }
+    Value new_op = rewriter.create<mhlo_disc::FakeQuantOp>(
+        loc, input.getType(), input, scale, zero_point, op.use_signedAttr(),
```
wyzero, line 668: why use input type instead of `op.getType()` as the result type of the new op.

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -643,6 +651,30 @@ LogicalResult ConvertDequantizeOp::dequantize(TF::DequantizeOp op,
...
+        op.use_symmetricAttr(), getI64ElementsAttr(axis, &rewriter),
+        op.num_bitsAttr(), op.quant_minAttr(), op.quant_maxAttr(),
+        op.use_dynamicAttr());
+    new_op.setType(op.output().getType());
```
wyzero, line 672: maybe we can remove this line after we changed line 668.

tao/tao_bridge/passes/tao_mark_for_compilation_pass.cc
qiuxiafei: done

tao/tao_bridge/ops/fake_quant_op.cc
qiuxiafei: done

tao_compiler/mlir/disc/transforms/lower_tf.cc
qiuxiafei: done

tao_compiler/mlir/disc/transforms/lower_tf.cc
qiuxiafei: done.

tao_compiler/mlir/disc/transforms/lower_tf.cc
qiuxiafei: done

PR #616, opened 2022-09-07
fix invalid concated shape in dot merge pass.

qiuxiafei:


scripts/python/tao_build.py
```diff
@@ -342,6 +342,8 @@ def _write(line, cmd="build"):
         _write("--host_jvm_args=-Djdk.http.auth.tunneling.disabledSchemes=", cmd = "startup")
         logger.info("configuring tao_bridge with bazel ......")
 
+    with cwd(root):
+        execute('rm -f tao_compiler/decoupling/version.h')
```
Orion34-lanbo, line 346: Why deleting version.h here?

wyzero: +1

qiuxiafei: if `version.h` exists (i.e. re-configure project), it will fail at L355, so delete it in advance.

qiuxiafei: anyway, I removed this explicit `rm -f` and use force copy on L354.

wyzero (approved):
LGTM

PR #638, opened 2022-09-19
[bugfix] using reverse transpose permutation in reverseIfOperandsAreConsistentTransposeOps

wyzero:
The PR fixed infinite conversions loops on optimizations of ResNets

### DIFF
--- a/tao_compiler/mlir/disc/transforms/disc_transpose_simplifier.cc
+++ b/tao_compiler/mlir/disc/transforms/disc_transpose_simplifier.cc
@@ -687,15 +687,15 @@ LogicalResult reverseIfOperandsAreConsistentTransposeOps(Operation* op,
 
   OpBuilder b(op);
   SmallVector<Value> newOperands;
-  newOperands.push_back(insertTranspose(op, lhs, permLHS, b)->getResult(0));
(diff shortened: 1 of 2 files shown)


fortianyou (approved):
LGTM

tao_compiler/mlir/disc/transforms/disc_transpose_simplifier.cc
```diff
@@ -687,15 +687,15 @@ LogicalResult reverseIfOperandsAreConsistentTransposeOps(Operation* op,
 
   OpBuilder b(op);
   SmallVector<Value> newOperands;
-  newOperands.push_back(insertTranspose(op, lhs, permLHS, b)->getResult(0));
-  newOperands.push_back(insertTranspose(op, rhs, permLHS, b)->getResult(0));
+  SmallVector<int64_t> reversePerm = getReversePermutation(permLHS);
```
qiuxiafei, line 690: So the bug origins from two consecutive TransposeOp what could not get eliminated?

wyzero: Yes. The permutation is wrong. Original UT is a special case where `permLHS (permLHS) = identity`, thus it's ok even the permutation is wrong. I added a new UT.

PR #639, opened 2022-09-19
take internal tao_compiler_main installation into consideration when create tf_blade package.

qiuxiafei:


tensorflow_blade/setup.py
```diff
@@ -129,6 +129,18 @@ def run(self):
                     os.symlink(fpath, link_name)
                     print(f"Link native lib: {fpath}")
 
+        # link internal disc.
+        import tensorflow as tf
```
fortianyou, line 133: Nit: Shall move to some prepare process in the internal package pipeline?

qiuxiafei: Not quite get your idea ... Actually we just forget to package `tao_compiler_main` into wheel, it is already there when running internal package pipeline.

fortianyou: My point is moving the modifications into the internal repo, but not this.

qiuxiafei: `tensorflow_blade's` whole project structure and all build scripts are placed on GitHub, in internal repo we only keep some C++ and Python source files, which are also linked `internal` directory under GitHub repo. There seems no place for this modification in internal repo..

fortianyou (approved):

PR #643, opened 2022-09-20
[Sparsity] Convert weight to sparse format in IR transform pass

yuchaoli:
Convert weight to sparse format in IR transform pass instead of runtime to reduce memory overhead.


tao_compiler/mlir/disc/IR/sparse_custom_call_op.cc
```diff
@@ -11,7 +11,7 @@ limitations under the License.
 ==============================================================================*/
 
 // This file defines sparse gemm related custom calls.
-
+#include "iostream"
```
wyzero, line 14: unused code?

yuchaoli: remove it.

PR #652, opened 2022-09-21
fix annotation bug and add some shape analysis support

zzpmiracle:


pytorch_blade/src/pybind.cpp
```diff
@@ -58,8 +59,11 @@ template <>
 void initModules<false>(py::module& m) {
   torch::blade::initToolsBindings(m);
   m.def(
-      "jit_pass_onnx_constant_f64_to_f32",
-      &torch::blade::CastDownAllConstantDoubleToFloat);
+       "jit_pass_onnx_constant_f64_to_f32",
+       &torch::blade::CastDownAllConstantDoubleToFloat)
+      .def(
+          "jit_pass_propagate_input_shapes",
+          &torch::blade::PropagateInputShapes);
```
fortianyou, line 66: Did you do clang-format?

pytorch_blade/torch_blade/exporter.py
```diff
@@ -45,6 +45,11 @@ def _set_annotate_args(s_module, annotations):
             continue
         input_dims, _ = annotations[idx-1]
         set_tensor_shape(input, input_dims)
+    # change input shape shape would not affect clusters input shape
+    # so propagate input shapes after setting annotations
+    from torch_blade import jit_pass_propagate_input_shapes
```
fortianyou, line 50: It's recommend to do import at the begining of the module.

pytorch_blade/src/compiler/jit/torch/shape_analysis.cpp
```diff
@@ -2209,14 +2214,20 @@ class ShapePropagator : public PropertyPropBase {
...
+        node->matches("aten::mul(Tensor self, Tensor other) -> Tensor") ||
+        node->matches("aten::mul_(Tensor self, Tensor other) -> Tensor")) {
       // These nodes handle tensors of different shapes internally, so there's
       // no need to insert explicit expand nodes.
       return PropagateShapeOnNodeByRunningIt(node);
-    } else if (node->matches(
-                   "aten::div(Tensor self, Tensor other) -> Tensor")) {
+    } else if (
+        node->matches("aten::div(Tensor self, Tensor other) -> Tensor") ||
+        node->matches("aten::div(Tensor self, Tensor other) -> Tensor")) {
```
fortianyou, line 2230: The two schemas are identical?

pytorch_blade/src/pybind.cpp
zzpmiracle: yes... `m` is too short to indent.

pytorch_blade/torch_blade/exporter.py
zzpmiracle: Ok.

PR #655, opened 2022-09-23
impl for lowering tf.Where

Orion34-lanbo:


tao_compiler/mlir/disc/transforms/lhlo_legalize_roots_to_loops.cc
```diff
@@ -4340,6 +4340,116 @@ LogicalResult lowerWithScheduleSparseSegmentMeanOpCPU(
...
+  // in llvm::CmpInst::CmpInst
+  if (input_elem_type.isa<IntegerType>()) {
+    Value zero_for_cmp =
+        b.create<arith::ConstantOp>(loc, b.getIntegerAttr(input_elem_type, 0));
+    is_zero = b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ne, input_elem,
+                                      zero_for_cmp);
+  } else if (input_elem_type.isa<FloatType>()) {
+    Value zero_for_cmp =
+        b.create<arith::ConstantOp>(loc, b.getFloatAttr(input_elem_type, 0));
+    is_zero = b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::ONE,
```
qiuxiafei, line 4415: what is `ONE` for...

qiuxiafei (commented):
on going...

tao_compiler/mlir/disc/transforms/lhlo_legalize_roots_to_loops.cc
Orion34-lanbo: According to mlir docs, `ONE` means `ordered not equal`，which is used for float type comparsion, however I am not quite sure which of `UNE` or `ONE` is more suitable(I don't quite get the exact meaning of these 2), after some trials `ONE` works the way I hope for this comparsion.

qiuxiafei: OK, got it ...  considering precision, floating type comparison maybe actually unstable. Just like writing normal C++ code, we just don't use `==` directly to compare two floating point numbers, instead, to check if they're close enough (like `np.isclose`).
So, does TF's WhereOp do something special on floating-point type?

qiuxiafei: See: https://floating-point-gui.de/errors/comparison/

tao_compiler/mlir/disc/IR/hlo_disc_ops.cc
```diff
@@ -699,6 +699,62 @@ LogicalResult SparseSegmentMeanOp::verify() {
...
+  Location loc = this->getLoc();
+  auto input_type = adaptor.input().getType().cast<RankedTensorType>();
+
+  SmallVector<Value, 2> index_shape_values, num_output_elements_shape_values;
+
+  Value idx_one = builder.create<arith::ConstantIndexOp>(loc, 1);
+  auto input_rank = input_type.getRank();
+  Value num_input_elements = idx_one;
+  for (int i = 0; i < input_rank; i++) {
+    Value dim_i = builder.create<tensor::DimOp>(loc, operands[0], i);
```
tanyokwok, line 719: Does it need an index cast here? Also, we prefer the bit-32 integer type in DISC. cc @wyzero

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -1342,6 +1342,65 @@ class ConvertSparseSegmentMeanOp
...
+        rewriter.create<tensor::ExtractOp>(loc, num_output_elements_vec,
+                                           idx_zero));
+    Value input_dims =
+        rewriter.create<tensor::DimOp>(loc, where.getResult(0), 1);
+
+    // output is rank 2
+    SmallVector<Value, 2> start_values(2, idx_zero);
+    SmallVector<Value, 2> limit_values(2, input_dims);
+    limit_values[0] = num_output_elements;
+    SmallVector<Value, 2> strides_values(2, idx_one);
```
tanyokwok, line 1371: Nit: I think the following codes are more human readable.
```
ArrayRef<Value> start_values{idx_zeros, idx_zeros};
ArrayRef<Value> limit_values{num_output_elements, input_dims};
ArrayRef<Value> strides_values{idx_one, idx_one):
```

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -1342,6 +1342,65 @@ class ConvertSparseSegmentMeanOp
...
+    auto limit_indices = rewriter.create<tensor::FromElementsOp>(
+        loc,
+        RankedTensorType::get({static_cast<int64_t>(limit_values.size())},
+                              index_ty),
+        limit_values);
+    auto strides_indices = rewriter.create<tensor::FromElementsOp>(
+        loc,
+        RankedTensorType::get({static_cast<int64_t>(strides_values.size())},
+                              index_ty),
+        strides_values);
```
tanyokwok, line 1387: Nit: use a lambda help function to refine those repeated codes.

tao_compiler/mlir/disc/transforms/lower_tf.cc
```diff
@@ -1342,6 +1342,65 @@ class ConvertSparseSegmentMeanOp
...
+    output_slice_shape_values.push_back(-1);
+    auto input_rank =
+        op.input().getType().dyn_cast<RankedTensorType>().getRank();
+    output_slice_shape_values.push_back(input_rank);
+
+    auto index_slice_op = rewriter.create<mhlo::RealDynamicSliceOp>(
+        loc,
+        RankedTensorType::get(output_slice_shape_values, rewriter.getI64Type()),
+        where.getResult(0), start_indices, limit_indices, strides_indices);
+    rewriter.replaceOp(op, index_slice_op.getResult());
```
tanyokwok, line 1399: Nit: It can be simplify into one single line.

`rewriter.replaceOpWithNewOp<mhlo::RealDynamicSliceOp>(op, op.getType(), where.getResult(0), start_indices, limit_indices, strides_indices);`

tao_compiler/mlir/disc/transforms/lhlo_legalize_roots_to_loops.cc
```diff
@@ -4340,6 +4340,116 @@ LogicalResult lowerWithScheduleSparseSegmentMeanOpCPU(
...
+      b.create<memref::LoadOp>(loc, num_output_elements, num_output_load_index);
+  Value output_index_casted =
+      b.create<arith::IndexCastOp>(loc, b.getIndexType(), output_index_count);
+  for (int i = 0; i < input_rank; ++i) {
+    llvm::SmallVector<Value, 2> store_index = {
+        output_index_casted, b.create<arith::ConstantIndexOp>(loc, i)};
+    b.create<memref::StoreOp>(loc, multidim_index_casted[i], index,
+                              store_index);
+  }
+  SmallVector<Value, 1>& num_output_store_index = num_output_load_index;
```
tanyokwok, line 4436: Nit: Change `SmallVector<Value, 1>&` to `ArrayRef<Value>`

tao_compiler/mlir/disc/transforms/lower_tf.cc
wyzero: It's not safe use ArrayRef here?

tao_compiler/mlir/disc/IR/hlo_disc_ops.cc
wyzero: it depends on the semantic of the op. The general rules are:

- use index type in `reifyReturnTypeShapes` if we know nothing about the range.
- use the same type as the shape operand of the op if we know it's safe.
- index type is ok since we can always lower it to i32 when lowering to llvm (because we know index type is always used for shape related calculation). i64 type is another story. it's not safe to    downcast it to i32 by default since i64 is not always used for shape.

tao_compiler/mlir/disc/transforms/lower_tf.cc
tanyokwok: I think they can be passed to the parameters directory, since they will not be modified.

wyzero: ArrayRef does not owned any container space. The temporary initialized list is destroyed after constructing the array ref.

tao_compiler/mlir/disc/transforms/lower_tf.cc
Orion34-lanbo: This will be modified by `clang-format` if in one single line ....

tao_compiler/mlir/disc/transforms/lhlo_legalize_roots_to_loops.cc
Orion34-lanbo: done

wyzero (approved):

PR #662, opened 2022-09-26
Delete old content from the disk remote cache

tanyokwok:
Thus we have to delete old caches, otherwise, we will run out of disk space. Because Bazel has no strategies to clean old caches. 
As Bazel documents suggested, this pull request manual deletes cache content exceeding 10 days from the local disk.
Ref: https://bazel.build/remote/caching#delete-remote-cache


Yancey1989 (approved):
LGTM++

PR #672, opened 2022-09-30
[WIP]refactor: use static init instead of call_once.

deeply:
Static var initialization is guaranteed to be called once by c++11 standard. Maybe we can remove some call_once usages for simplicity (and maybe small perf gain).

### DIFF
--- a/pytorch_blade/src/compiler/mlir/converters/torch_mlir_op_filter.cpp
+++ b/pytorch_blade/src/compiler/mlir/converters/torch_mlir_op_filter.cpp
@@ -37,8 +37,8 @@ bool IsTorchMlirSupported(const torch::jit::Node& node) {
 }
 
 // clang-format off
-const std::unordered_set<std::string> &GetTorchMlirWhiteList() {
-  static std::unordered_set<std::string> white_list{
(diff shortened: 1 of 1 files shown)


pytorch_blade/src/compiler/mlir/converters/torch_mlir_op_filter.cpp
```diff
@@ -113,24 +113,26 @@ const std::unordered_set<std::string> &GetTorchMlirWhiteList() {
...
-      }
-      ostr << "]";
-      LOG(INFO) << ostr.str();
-  });
   return white_list;
 }
 // clang-format off
 
+const std::unordered_set<std::string> &GetTorchMlirWhiteList() {
+  static std::unordered_set<std::string> white_list = CreateTorchMlirWhiteList();
```
tanyokwok, line 133: Nit: Could you please link the c++ 11 standard doc reference?

deeply: please refer to this [SO answer](https://stackoverflow.com/a/1661564) or [C++11 standard](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf) 6.7

PR #676, opened 2022-10-10
fix #462, Remove target 'torch_blade_include`

### LINKED ISSUE #462: torch-blade's `//src:torch_blade_include` may lead to illegal cache behaviour.
https://github.com/alibaba/BladeDISC/blob/336736c89b4ccc2a5da28c1c43afd85be5e72893/pytorch_blade/src/BUILD#L15-L21

By depending on the above, bazel targets may have C++ source root as include root other than bazel workspace root. As a consequence, targets my reference any header file without explicitly adding it to `deps`. This may cause some problem when remote build cache is introduced. Bazel seems not treating change of any header file under `.` as change of target `torch_blade_include`. 

It's better to have explicit header dependency.

qiuxiafei:
Remove target 'torch_blade_include` and rename directory 'src` to 'pytorch_blade`.


pytorch_blade/pytorch_blade/BUILD
```diff
@@ -30,19 +30,19 @@ cc_library(
     name = "torch_blade",
     linkopts = ["-Wl,-rpath,$$ORIGIN"],
     deps = [
-        "//src/common_utils:torch_blade_common",
-        "//src/compiler/backends:torch_blade_backends",
-        "//src/compiler/jit:torch_blade_jit",
-        "//src/compiler/mlir:torch_blade_mlir",
+        "//pytorch_blade/common_utils:torch_blade_common",
```
tanyokwok, line 33: Nit: remove the target `torch_blade_include` at line 22

pytorch_blade/bazel_build.py
```diff
@@ -45,8 +45,8 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.targets = [
             "@org_tensorflow//tensorflow/compiler/mlir/xla/ral:libral_base_context.so",
-            "//src:libtorch_blade.so",
-            "//src:_torch_blade.so",
+            "//pytorch_blade:libtorch_blade.so",
```
tanyokwok, line 48: Nit：Please update the targets in the file `distribution.cfg` as well

pytorch_blade/pytorch_blade/BUILD
qiuxiafei: done~

tanyokwok (approved):
LGTM

PR #678, opened 2022-10-10
Limit make job number to reduce memory footprint

qiuxiafei:


third_party/bazel/third_party/blade_gemm.BUILD
```diff
@@ -32,7 +32,7 @@ cmake(
         "CUDACXX": blade_gemm_nvcc(),
     },
     generate_crosstool_file=False, ## This makes sure we use cxx by cache_entries settings
-    build_args = ["-j"],
+    build_args = ["-j", "32"],  # Limit make job number to reduce memory usage.
```
tanyokwok, line 35: Nit: If there have CPU cores lower than 33, it will run into dead cycles. I think it's better to use `min(32, $nproc - 1)`.

tanyokwok (approved):
LGTM

PR #683, opened 2022-10-13
[quantization] Support qgemm int8 for BladeDISC on GPU

linbinskn:
to #567 
- [x] const weight transpose pass
- [x] bladnn int8 gemm library call
- [x] end2end subgraph test (only work when blade_gemm is on)
- [x] qconv related bug fix

### DIFF
--- a/tao_compiler/mlir/disc/transforms/quantized_dot_rewriter.cc
+++ b/tao_compiler/mlir/disc/transforms/quantized_dot_rewriter.cc
@@ -94,15 +94,24 @@ struct QuantDotTransposeConvert
       return failure();
     }
 
-    if (op.weight().getDefiningOp<mhlo::TransposeOp>()) {
+    auto inputTy = op.input().getType().dyn_cast<RankedTensorType>();
+    auto weightTy = op.weight().getType().dyn_cast<RankedTensorType>();
+
+    // This transpose pass can only support dot with rank equal to 2
+    if (inputTy.getRank() != 2 || weightTy.getRank() != 2) {
       return failure();
     }
 
-    auto weightTy = op.weight().getType().dyn_cast<RankedTensorType>();
-    int rank = weightTy.getRank();
+    auto dim_numbers = op.dot_dimension_numbers();
+    SmallVector<int64_t, 4> rhs_perm_check;
+    auto rhs_batching_dims_check = dim_numbers.getRhsBatchingDimensions();
+    bool tp_rhs_check = isNonBatchingTransposeTensorValue(
+        op.weight(), rhs_perm_check,
+        std::unordered_set<int64_t>(rhs_batching_dims_check.begin(),
+                                    rhs_batching_dims_check.end()));
 
-    // This transpose pass can only support dot with rank equal to 2
-    if (rank != 2) {
+    // Match fail if weight is nxk
+    if (tp_rhs_check) {
       return failure();
     }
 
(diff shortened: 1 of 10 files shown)


tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -401,6 +404,64 @@ void ral_gemm(ExecutionContext* ctx, void* stream_handle, MemRefType<InT, 2> A,
...
+  TAO_VLOG(0) << "weight.sizes[0]: " << weight.sizes[0]
+              << " weight.sizes[1]: " << weight.sizes[1];
+  TAO_VLOG(0) << "result.sizes[0]: " << result.sizes[0]
+              << " result.sizes[1]: " << result.sizes[1];
+
+  auto gpu_driver = ctx->getDriver<GPUDriver>(GPUDriver::name());
+  float input_scale, weight_scale, result_scale;
+  gpu_driver->d2h(ctx, stream_handle, &inputScales.data[0], &input_scale,
+                  sizeof(float));
+  gpu_driver->d2h(ctx, stream_handle, &weightScales.data[0], &weight_scale,
```
wyzero, line 431: I see the UT using per-tensor quantization  for weight while we only handle the first element of the scale. Is it the supposed behavior?

tao_compiler/mlir/disc/transforms/quantized_dot_rewriter.cc
```diff
@@ -0,0 +1,209 @@
...
+
+  LogicalResult matchAndRewrite(mlir::mhlo_disc::QuantizedDotGeneralOp op,
+                                PatternRewriter& rewriter) const override {
+    bool onGpu = placement_utils::isGpuMhlo(op);
+    // Only need to transpose weight on gpu currently
+    if (onGpu == false) {
+      return failure();
+    }
+
+    if (op.weight().getDefiningOp<mhlo::TransposeOp>()) {
```
wyzero, line 97: The candidate `QuantizedDotGeneralOp` ops are those having `k*n` layout, not those having `TransposeOp` operands. It's not reliable to only check the existence of the transpose op (e.g. transpose + dot op having n x k layout should also be taken into consideration).

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
linbinskn: Yes, we only support per tensor quantization on gpu. This ut is used for per tensor testing although its weight scale is per channel.

tao_compiler/mlir/disc/transforms/quantized_dot_rewriter.cc
```diff
@@ -0,0 +1,209 @@
...
+      return failure();
+    }
+
+    auto weightTy = op.weight().getType().dyn_cast<RankedTensorType>();
+    int rank = weightTy.getRank();
+
+    // This transpose pass can only support dot with rank equal to 2
+    if (rank != 2) {
```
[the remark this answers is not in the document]
linbinskn: Done

tao_compiler/mlir/disc/transforms/quantized_dot_rewriter.cc
linbinskn: Make sense, done.

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
wyzero: Then why not use a per-tensor UT directly？the compiler side is supposed to be ok for both cases. And the register kernel should use `memref<f32>` for weight scale/zero_point inputs.

linbinskn: Of course we can use a per-tensor UT directly. However, I think per-channel/per-tensor should be unified in library call interface although we may launch different kernels for them. And if we support per-channel quantization later, we only need to modify ral kernel library call function.

wyzero (approved):

PR #684, opened 2022-10-13
[cpu] add a knob for GEMM weight pre-packing

wyzero:
Add a new env to control weight pre-packing for GEMM ops. This enables more fine-grained configuration.

### DIFF
--- a/tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
+++ b/tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
@@ -137,6 +137,18 @@ bool isWeightPrePackingEnabled() {
   return enabled;
 }
 
+bool isWeightPrePackingForMatMulEnabled() {
+  static bool enabled = []() {
+    const char* env = getenv("DISC_CPU_ENABLE_WEIGHT_PRE_PACKING_FOR_MATMUL");
+    if (!env) return true;
+    std::string envStr = env;
+    std::transform(envStr.begin(), envStr.end(), envStr.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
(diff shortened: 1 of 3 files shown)


tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
```diff
@@ -137,6 +137,18 @@ bool isWeightPrePackingEnabled() {
...
+bool isWeightPrePackingForMatMulEnabled() {
+  static bool enabled = []() {
+    const char* env = getenv("DISC_CPU_ENABLE_WEIGHT_PRE_PACKING_FOR_MATMUL");
+    if (!env) return true;
+    std::string envStr = env;
+    std::transform(envStr.begin(), envStr.end(), envStr.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
+    return envStr == "true" || envStr == "1";
+  }();
+  return isWeightPrePackingEnabled() && enabled;
```
minminsun, line 149: Seems that this is for some senario where pre_packing for mamtul should be disabled while pre_packing for others (mostly conv) should be enabled?

wyzero: Yes. We rely on onednn to call ACL gemm while we call acl conv directly currently. Thus, we can only share packed weight across different shapes for conv op at the moment. We'll fix this later by bypassing onednn.

minminsun (approved):

tao_compiler/mlir/xla/ral/context/common_context_impl_mkldnn.cc
minminsun: I see. Thanks!

PR #687, opened 2022-10-14
[AArch64] re-use ACL depthwise operator instance for different threads

wyzero:
Depthwise conv op has been refactored to support parallel `op.run...()`. This CL changes accordingly to enable such feature.


qiuxiafei (approved):
LGTM

PR #690, opened 2022-10-17
[AArch64] re-use ACL conv operator instance for different threads

wyzero:
1, patch ACL to support parallel `conv.run(...)`
2, change DISC accordingly to enable the above feature.


qiuxiafei (approved):
LGTM

Issue #692, opened 2022-10-19
The pytorch demo for bert-base-multilingual-uncased-sentiment does not work.


LiZerun:
I just follow the instructions on https://github.com/alibaba/BladeDISC/blob/main/docs/tutorials/torch_bert_inference.md.
the docker file I use is: **bladedisc/bladedisc:latest-runtime-torch1.12.0-cu113**
and the GPU is **A100**
When trying to optimize the module with torchblade, it gives the following error info:

```
[E mhlo_conversion.cpp:261]
bad optional access
The above operation failed shape propagation in this context:
/usr/local/lib/python3.8/dist-packages/transformers/models/bert/modeling_bert.py(217): forward
/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py(1118): _slow_forward
/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py(1130): _call_impl
/usr/local/lib/python3.8/dist-packages/transformers/models/bert/modeling_bert.py(1007): forward
/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py(1118): _slow_forward
/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py(1130): _call_impl
/usr/local/lib/python3.8/dist-packages/transformers/models/bert/modeling_bert.py(1552): forward
/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py(1118): _slow_forward
/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py(1130): _call_impl
/usr/local/lib/python3.8/dist-packages/torch/jit/_trace.py(967): trace_module
/usr/local/lib/python3.8/dist-packages/torch/jit/_trace.py(750): trace
/usr/local/lib/python3.8/dist-packages/torch_blade/exporter.py(253): export
/usr/local/lib/python3.8/dist-packages/torch/autograd/grad_mode.py(27): decorate_context
/usr/local/lib/python3.8/dist-packages/torch_blade/optimization.py(36): _optimize
/usr/local/lib/python3.8/dist-packages/torch/autograd/grad_mode.py(27): decorate_context
/usr/local/lib/python3.8/dist-packages/torch_blade/optimization.py(108): optimize
/usr/local/lib/python3.8/dist-packages/torch/autograd/grad_mode.py(27): decorate_context
test.py(52): <module>
```

tanyokwok:
[...]

PR #694, opened 2022-10-20
fix aten::slice shape propogation

tanyokwok:
The PR should fix the bug found in issue #692.


pytorch_blade/pytorch_blade/compiler/jit/torch/shape_analysis.cpp
```diff
@@ -2438,36 +2442,6 @@ class ShapePropagator : public PropertyPropBase {
         node->output()->setType(type->withSymbolicShapes(newSizes));
       }
       return true;
-    } else if (
```
wyzero, line 2441: Why delete this part directly? what's the difference between this part and the above part?

tanyokwok: This function will use input complete shape info to infer the node's output shape. For `aten::slice`, they make no difference here.

wyzero (approved):

PR #696, opened 2022-10-21
Enable fully dynamic shapes in inference

tanyokwok:
After switching the converter to Torch-MLIR, we must erase the recorded concrete shapes on the graph inputs. Otherwise, the shapes will be regarded as input information and used directly by the shape inference and constraints.


Yancey1989 (approved):
LGTM

PR #700, opened 2022-10-21
fix pybind11 abi

tanyokwok:
To fix errors like when input with correct arguments:
```
TypeError: graph_create_get_attr(): incompatible function arguments. The following argument types are supported:
    1. (arg0: torch::jit::Graph, arg1: torch::jit::Value, arg2: str) -> torch::jit::Node
```

### DIFF
--- a/pytorch_blade/bazel_build.py
+++ b/pytorch_blade/bazel_build.py
@@ -14,6 +14,7 @@ import os
 import errno
 import subprocess
 import sys
+import torch
 import venv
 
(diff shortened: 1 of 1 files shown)


tanyokwok:
> 

pybind11 has the internals struct shared between all the modules, but the module compiled from different ABI will be queried with different PYBIND11_INTERNALS_ID, see https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/internals.h#L422

qiuxiafei (approved):

PR #703, opened 2022-10-26
 [PDLL] torch custom call plugin tool based on mlir pdll

wyzero:


tanyokwok (approved):
LGTM!

I think we should be more tolerant of new trial usage of the extension.

PR #708, opened 2022-10-27
support clustering for cpu sparse ops

Orion34-lanbo:


tao/tao_bridge/tf/xla_op_registry.cc
```diff
@@ -1244,9 +1244,13 @@ REGISTER_XLA_OP_FOR_TAO(Name("Dequantize"));
 #if defined(TAO_AARCH64)
 // TODO(disc): support `QuantizedConv2DWithBiasAndRequantize` on other platforms
 REGISTER_XLA_OP_FOR_TAO(Name("QuantizedConv2DWithBiasAndRequantize"));
-#endif
-#endif
-#endif
+#endif // TAO_AARCH64
```
qiuxiafei, line 1247: we may have it enabled on arm right?

Orion34-lanbo: It is not fully tested on `aarch64` yet, so not enabled.

qiuxiafei (approved):

PR #709, opened 2022-10-28
Add pre-built torch into ci

tanyokwok:
This PR adds a nightly preview torch version into CI. It fixes and updates the following things:

1. fix apt sources establishment in Dockerfile
2. fix compilation errors introduced by API changes of Torch


pytorch_blade/pytorch_blade/compiler/jit/shape_type_spec.cpp
```diff
@@ -97,11 +97,11 @@ ShapeType ShapeType::Deserialize(const std::string& serial_str) {
         // string like: 1:1 (min:max)
         CHECK(!s.empty());
         std::vector<std::string> dim_range = split(s, ":");
-        CHECK_EQ(dim_range.size(), 2);
+        CHECK(dim_range.size() == 2);
```
qiuxiafei, line 100: `CHECK` and `TORCH_CHECK` are both used in this file ...  so where is `CHECK` from ... and why `CHECK` is used...

qiuxiafei (approved):

PR #715, opened 2022-11-01
Change download script for mkl

tanyokwok:
Try a stable choice of downloading and extracting tar.

### DIFF
--- a/scripts/python/common_setup.py
+++ b/scripts/python/common_setup.py
@@ -272,8 +272,12 @@ def config_mkldnn(root, args):
             # download mkl-lib/include
             download_cmd = """
               unset HTTPS_PROXY
-              curl -fsSL https://pai-blade.oss-accelerate.aliyuncs.com/build_deps/mkl/mkl-static-2022.0.1-intel_117.tar.bz2 | tar xjv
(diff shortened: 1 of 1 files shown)


scripts/python/common_setup.py
```diff
@@ -272,8 +272,12 @@ def config_mkldnn(root, args):
             # download mkl-lib/include
             download_cmd = """
               unset HTTPS_PROXY
-              curl -fsSL https://pai-blade.oss-accelerate.aliyuncs.com/build_deps/mkl/mkl-static-2022.0.1-intel_117.tar.bz2 | tar xjv
-              curl -fsSL https://pai-blade.oss-accelerate.aliyuncs.com/build_deps/mkl/mkl-include-2022.0.1-h8d4b97c_803.tar.bz2 | tar xjv
+              wget -nv https://pai-blade.oss-accelerate.aliyuncs.com/build_deps/mkl/mkl-static-2022.0.1-intel_117.tar.bz2 -O mkl-static-2022.0.1-intel_117.tar.bz2
```
qiuxiafei, line 275: How could this be more stable?

tanyokwok: Just try and run. Indeed, the former script use unix pipeline as inputs of tar could fragile if the network is not stable.

qiuxiafei (approved):

PR #750, opened 2022-11-15
[PDLL] Add torch qgemm pattern match

linbinskn:
dequant+gemm+quant -> qgemm
Parsed Payload module:
image
Rewrited Payload module:
image


wyzero (approved):
LGTM

PR #752, opened 2022-11-15
[build] make torch build script hookable from internal

wyzero:


qiuxiafei (approved):
pls also have a try on internal build.

PR #781, opened 2022-11-23
[PDL]enable pdl pattern match in mlir

chenbohua3:


pytorch_blade/pytorch_blade/torch-mlir/lib/utils/env.cpp
```diff
@@ -0,0 +1,60 @@
...
+// limitations under the License.
+
+#include "utils/env.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace mlir {
+namespace torch {
+namespace utils {
+namespace env {
+std::string AsciiStrToLower(const char* cstr) {
```
qiuxiafei, line 19: UT?

pytorch_blade/pytorch_blade/torch-mlir/lib/Conversion/MhloPasses.cpp
```diff
@@ -15,6 +15,9 @@
 #include "torch-mlir/Conversion/TorchToSCF/TorchToSCF.h"
 #include "torch-mlir/Dialect/Torch/Transforms/Passes.h"
 #include "torch-mlir/Dialect/TorchConversion/Transforms/Passes.h"
+#include "utils/env.h"
+
+//#include "pytorch_blade/common_utils/env.h"
```
wyzero, line 20: remove unused code.

chenbohua3: removed

pytorch_blade/pytorch_blade/torch-mlir/lib/utils/env.cpp
chenbohua3: I just copied these functions from pytorch_blade/common_utils. Had added a comment about this.

pytorch_blade/tests/disc/pdl/__init__.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2021 The BladeDISC Authors. All rights reserved.
```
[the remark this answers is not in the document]
chenbohua3: As described in doc of unittest: Start directory and subdirectories containing tests for unittest must be regular package that have __init__.py file.

pytorch_blade/pytorch_blade/torch-mlir/lib/utils/env.cpp
chenbohua3: log utils in pytorch_blade depend on PyTorch log system which will introduce PyTorch as a dependence. Furthermore, PyTorch depends on llvm which will be conflict with that in torch-mlir.

pytorch_blade/pytorch_blade/torch-mlir/lib/Conversion/MhloPasses.cpp
```diff
@@ -42,6 +43,17 @@ void mlir::torch::createDiscTorchBackendToMhloBackendPipeline(
   funcOptions.decompose = false;
   ::mlir::torch::createDiscTorchFunctionToTorchBackendPipeline(pm, funcOptions);
 
+  // Apply pdl pattern match
+  std::string disc_torch_pdl_files;
+  std::string disc_torch_pdll_include_dirs;
+  disc_torch_pdl_files =
```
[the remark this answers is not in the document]
chenbohua3: To be consistent with the behavior in tf. Specifically, since pdl is still an experimental feature, it is not suitable to enable it by default.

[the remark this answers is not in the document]
wyzero: Currently we support two different mode:

1,  pre-defined pdl patterns. These patterns will be compiled with the disc_compiler_main binary.

2,  support read outside pdl patterns at runtime. This mode is also useful since it enables us to extend DISC without re-building it.

We can move a pattern to the pre-defined set once we believe it becomes more stable.

wyzero (approved):

PR #805, opened 2022-11-28
[Quantization] change cast type to avoid h2d/d2h

linbinskn:
- [x] avoid extra h2d/d2h by casting zeropoint to float 
- [x] bug fix of convergence failure in quantized dot rewriter
- [x] bug fix in ral bladnn kernel call


tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -1587,7 +1587,7 @@ bool ral_conv_bladnn(ExecutionContext* ctx, void* stream_handle,
     ki = kernel.sizes[3];
     oh = output.sizes[1];
     ow = output.sizes[2];
-    oc == output.sizes[3];
+    oc = output.sizes[3];
```
wyzero, line 1590: nit: it's not a bug of stream executor. maybe refine the description of your PR.

tao_compiler/mlir/disc/transforms/disc_lower_quantize_and_dequantize.cc
```diff
@@ -119,11 +119,12 @@ struct DequantizeOpConverter
...
-    auto bcastedInputOrZeroPointTy =
-        RankedTensorType::get(inputTy.getShape(), zeroPointTy.getElementType(),
-                              zeroPointTy.getEncoding());
+    auto bcastedInputOrZeroPointTy = RankedTensorType::get(
+        inputTy.getShape(), scaleTy.getElementType(), scaleTy.getEncoding());
+    Value castedZeroPoint =
+        rewriter.create<mhlo::ConvertOp>(loc, scaleTy, op.getZeroPoint());
     Value bcastedZeroPoint = rewriter.create<mhlo::DynamicBroadcastInDimOp>(
-        loc, bcastedInputOrZeroPointTy, op.getZeroPoint(), inputShape,
+        loc, bcastedInputOrZeroPointTy, castedZeroPoint, inputShape,
```
wyzero, line 127: I'm not quite sure if it's always safe to cast` input/zero_point `to float before we do `input-zeropoint`. What if the quantization is int32 instead of int8? maybe we can leave a comment here to describe the requirements of such change.

tao_compiler/mlir/disc/transforms/quantized_dot_rewriter.cc
```diff
@@ -102,16 +102,11 @@ struct QuantDotTransposeConvert
...
-        op.getWeight(), rhs_perm_check,
-        std::unordered_set<int64_t>(rhs_batching_dims_check.begin(),
-                                    rhs_batching_dims_check.end()));
-
     // Match fail if weight is nxk
-    if (tp_rhs_check) {
+    std::vector<int64_t> rhs_contracting_dims_check;
+    auto dim_numbers = op.getDotDimensionNumbers();
+    rhs_contracting_dims_check = dim_numbers.getRhsContractingDimensions();
+    if (rhs_contracting_dims_check[0] == 1) {
```
wyzero, line 109: what if `getRhsContractingDimensions` is `[0, 1]`? e.g. rank-2 vector x rank2 vector -> scalar?

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
linbinskn: Done

tao_compiler/mlir/disc/transforms/quantized_dot_rewriter.cc
linbinskn: This rewrite function only match ContractingDimensions.size() == 1, have added constraint.

tao_compiler/mlir/disc/transforms/disc_lower_quantize_and_dequantize.cc
linbinskn: In my point of view, float32 would be enough in most cases. Although this modification may bring risk, it can reduce large overhead meanwhile. Have added comments in case.

wyzero (dismissed):

wyzero (approved):

PR #807, opened 2022-11-28
refinement for internal build about `blade gemm` and `compute intensive fusion`

Orion34-lanbo:
blade gemm is not needed for `compiler` side build, thus remove related build logic.

### DIFF
--- a/scripts/python/tao_build.py
+++ b/scripts/python/tao_build.py
@@ -133,8 +133,6 @@ def restore_gcc_conf(args):
 
 
 def configure_compiler(root, args):
-    config_blade_gemm(root, args)
-
     def _action_env(key, value, cmd="build"):
         f.write(f"{cmd} --action_env {key}={value}\n")
     # configure tensorflow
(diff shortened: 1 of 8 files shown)


minminsun (dismissed):

scripts/python/tao_build.py
```diff
@@ -170,35 +168,6 @@ def _action_env(key, value, cmd="build"):
     configure_compiler_platform_alibaba(root, args)
     logger.info("Stage [configure compiler] success.")
 
-def config_blade_gemm(root, args):
```
qiuxiafei, line 173: just ensure blade_gemm built here is never used ?

qiuxiafei (approved):
LGTM with minor comments.

scripts/python/tao_build.py
Orion34-lanbo: This is for `tao_compiler` build, I have checked with @minminsun this is no not needed.

PR #813, opened 2022-11-29
[Quantization] Add bias quant

linbinskn:
The target of this pr is to be compatible with int8 kernel implementation, namely int32 biasadd for cpu and int8 biasadd for gpu.

### DIFF
--- a/pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
+++ b/pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
@@ -63,8 +63,6 @@ class TestLinear(CPUDiscPdlQuantizationTestCase):
                     self.activation_quant_min, self.activation_quant_max
                 )
                 return x
-        if self.device == torch.device('cuda'):
-            return
         model = Model().eval().to(self.device)
         inp = torch.randn(1, 2, 128).to(self.device)
         traced_model = torch.jit.trace(model, inp)
--- a/pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
+++ b/pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
@@ -1,5 +1,3 @@
-
-
 # Copyright 2021 The BladeDISC Authors. All rights reserved.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
+++ b/pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
@@ -159,8 +159,6 @@ class TestGPULinear(GPUDiscPdlQuantizationTestCase):
                     self.activation_quant_min, self.activation_quant_max
                 )
                 return x
-        if self.device != torch.device('cuda'):
-            return
         model = Model().eval().to(self.device)
         inp = torch.randn(1, 2, 128).to(self.device)
         traced_model = torch.jit.trace(model, inp)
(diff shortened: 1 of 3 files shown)


chenbohua3:
I think it is better to construct a stand-alone pdll file to support bias with fake-quant, since it is not a common way.

linbinskn:
> I think it is better to construct a stand-alone pdll file to support bias with fake-quant, since it is not a common way.

Only add gpu bias quant this pr. Final solution will be discussed later.

pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
```diff
@@ -63,7 +63,8 @@ def forward(self, x):
                     self.activation_quant_min, self.activation_quant_max
                 )
                 return x
-
+        if self.device == torch.device('cuda'):
```
chenbohua3, line 66: It is better to build different base classes for different device (e.g. gpu and cpu) to enable testing on the respective hardware. Also, skip test using `return` may not be a good idea, use the api of `unitest` instead

qiuxiafei:
Just adding tests?

linbinskn:
> Just adding tests?

In addition to add test, it also modifies and adds pdll files.

pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
linbinskn: Have rebased. Also add the following rewrite pattern for gpu kernel implementation. The equality between original graph and rewritten graph is not strict. 
    #  weight -> fake-quant  \
    #  input  -> fake-quant -> linear -> fake-quant -> output
    #                   bias /
    #
    # rewrited graph:
    #
    #  weight -> quant  \
    #  input  -> quant -> qgemm -> dequant -> output
    #  bias   -> quant  /

wyzero (dismissed):

wyzero (dismissed):

pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
chenbohua3: remove this codes?

linbinskn: done

pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
```diff
@@ -1,3 +1,5 @@
+
```
chenbohua3, line 1: remove these empty lines

pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
```diff
@@ -294,10 +488,9 @@ def forward(self, x):
...
   }
 }
-
         """
         self._test_torchscipte_to_mhlo(traced_model._c, expect_str, self.pdll_files, enable_int8=True)
 
 
 if __name__ == "__main__":
-    unittest.main()
+    unittest.main()
```
chenbohua3, line 496: Although we do not enable python style check in this repo, it is better for each developer to obey the pep8 rules. There should be an empty line in the end of each python file

pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
linbinskn: Done.

pytorch_blade/tests/disc/pdl/test_torchscipt_to_mhlo/test_quantization.py
linbinskn: Agree with you, have modified.

chenbohua3 (approved):

PR #819, opened 2022-11-29
[Quantization][PDL] pdll qgemm kernel for x86 and a e2e test

chenbohua3:
1. a pdl qgemm kernel with bias support (rank=2 for operand 1)
2. a e2e test case for x86 platform with acc check

### DIFF
--- a/tao_compiler/mlir/xla/ral/context/common_context_impl_quantization.cc
+++ b/tao_compiler/mlir/xla/ral/context/common_context_impl_quantization.cc
@@ -10,7 +10,6 @@
 // limitations under the License.
 
 #if defined(TAO_CPU_ONLY)
-
 #include <thread>
 
 #include "tensorflow/compiler/mlir/xla/ral/context/common_context_impl_mkldnn.h"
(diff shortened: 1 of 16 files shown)


tao_compiler/mlir/xla/ral/context/common_context_impl_quantization.cc
```diff
@@ -726,6 +725,102 @@ void ral_qgemm_onednn_s8_s8_s8_per_channel(
...
+  ideep::tensor input_t{dims{m, k}, ideep::data_type::s8,
+                        tp_a ? format_tag::ba : format_tag::ab, input.data};
+  ideep::tensor weight_t{dims{k, n}, ideep::data_type::s8,
+                         tp_b ? format_tag::ba : format_tag::ab, weight.data};
+  ideep::tensor bias_t{dims{1, n}, ideep::data_type::f32, format_tag::ab,
+                       bias.data};
+  ideep::tensor output_t{dims{m, n}, ideep::data_type::s8, format_tag::ab,
+                         result.data};
+
+  std::vector<float> input_scales({inputScales.data[0]});
```
wyzero, line 791: If not online quantization, scale/zero_points are all constants. One possible optimization is to make these settings as attributes instead of inputs of the kernel. And then `input_scales ` and all other similar settings can be cached to reduce overhead. This can also help to reduce const op dispatch overhead in DISC. Current implementation is ok for PR and we may choose to do it in the future.

tao_compiler/mlir/xla/ral/context/common_context_impl_quantization.cc
```diff
@@ -726,6 +725,102 @@ void ral_qgemm_onednn_s8_s8_s8_per_channel(
   timer.Stop();
 }
 
+MemRefType<int8_t, 2> ral_pdll_qgemm_onednn_s8_s8_s8_per_channel(
+    ExecutionContext* ctx, opaque_t /*stream_handle*/,
+    MemRefType<int8_t, 2> input, MemRefType<int8_t, 2> weight,
+    MemRefType<float, 1> bias, MemRefType<float, 0> inputScales,
```
wyzero, line 731: so the bias input is not quantized as other inputs?

tao_compiler/mlir/xla/ral/context/common_context_impl_quantization.cc
chenbohua3: Agree, I will optimize the implementation in the future pr.

tao_compiler/mlir/xla/ral/context/common_context_impl_quantization.cc
chenbohua3: yes, leave it to oneDNN to convert it s32 using the scale we passed to

wyzero (approved):

PR #820, opened 2022-11-30
fix build with deeprec-gpu

Orion34-lanbo:


tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -1392,6 +1392,9 @@ Status RunCudnnConvolution<int8_t>(
     se::DeviceMemoryBase& result_buffer,
     se::ScratchAllocator* scratch_allocator, se::Stream* stream,
     se::dnn::ProfileResult* profile_result) {
+#if defined(TF_IS_DEEPREC)
```
qiuxiafei, line 1395: DeepRec doesn't support `RunCudnnConvolution` ?

Orion34-lanbo: DeepREC is tf-1.15, which does not support `RunCudnnConvolution` with `int8_t` type.

PR #833, opened 2022-12-05
Fix getting wrong source root path when building pytorch_blade

bikekiller:
pytorch_blade created a symbolic link to common_setup.py which use os.path.abspath() to get absolute path for itself. This make get_source_root_dir() return incorrect path to the BladeDISC source root when building pytorch_blade.

### DIFF
--- a/scripts/python/common_setup.py
+++ b/scripts/python/common_setup.py
@@ -536,8 +536,12 @@ def internal_tao_bridge_dir():
 def link_dirs(dst_dir, src_dir):
     execute("rm -rf {0} && ln -s {1} {0}".format(dst_dir, src_dir))
 
-def try_import_from_platform_alibaba(module_name):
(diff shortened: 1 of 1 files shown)


scripts/python/common_setup.py
```diff
@@ -536,8 +536,12 @@ def internal_tao_bridge_dir():
 def link_dirs(dst_dir, src_dir):
     execute("rm -rf {0} && ln -s {1} {0}".format(dst_dir, src_dir))
 
-def try_import_from_platform_alibaba(module_name):
-    build_scripts_dir = os.path.join(internal_root_dir(), "platform_alibaba", "scripts", "python")
+def try_import_from_platform_alibaba(module_name, disc_root=None):
+    if disc_root is None:
+        root = internal_root_dir()
+    else:
+        root = os.path.join(disc_root, os.path.pardir)
```
wyzero, line 543: why not use disc_root directly?

wyzero (approved):
LGTM. thanks~

PR #835, opened 2022-12-05
[Quantization] qgemm ral library call and subgraph e2e test

linbinskn:


tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -1869,6 +1928,10 @@ TAO_RAL_API("ral_conv", "gpu",
             gpu::se_impl::gpu_conv_impl::ral_conv<Eigen::half, 3>);
 TAO_RAL_API("ral_qconv", "gpu",
             gpu::se_impl::gpu_conv_impl::ral_qconv<int8_t, 4>);
+TAO_RAL_API("ral_pdll_qgemm", "gpu",
+            gpu::se_impl::ral_pdll_qgemm_s8s8s8_biasadd_quant_per_tensor<2>);
+TAO_RAL_API("ral_pdll_qgemm", "gpu",
+            gpu::se_impl::ral_pdll_qgemm_s8s8s8_biasadd_quant_per_tensor<3>);
```
minminsun, line 1934: `ral_gemm` only supports 2D tensors, why does qgemm need to support 3D tensors?

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -470,6 +470,65 @@ void ral_qgemm(
...
+  result_scale = resultScales.data[0];
+
+  float kernel_scale = input_scale * weight_scale / result_scale;
+  float beta = 1.0f;
+
+#if defined(PLATFORM_ALIBABA) and defined(ENABLE_BLADE_GEMM)
+  {
+    auto gpu_driver = ctx->getDriver<GPUDriver>(GPUDriver::name());
+    auto data =
+        static_cast<int8_t*>(gpu_driver->alloc(ctx, m * n * sizeof(int8_t)));
```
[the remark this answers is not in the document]
linbinskn: Buffer has to be allocated in ral gemm function because pdll based customcall doesn't know the output shape in current implementation.

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
wyzero: If we still only want to handle the 2d case, we can insert a reshape op in the PDL file manually.

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -470,6 +470,65 @@ void ral_qgemm(
...
+  for (int i = 0; i < N - 1; i += 1) {
+    m *= input.sizes[i];
+    resultSizes[i] = input.sizes[i];
+  }
+  resultSizes[N - 1] = n;
+
+  if (isEmptyMemref(input) || isEmptyMemref(weight)) {
+    TAO_VLOG(1) << "ral_qgemm: early return for empty tensor";
+    return assignMemRef<int8_t, N>(nullptr, resultSizes);
+    ;
```
wyzero, line 494: remove unused code

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -470,6 +470,65 @@ void ral_qgemm(
   ctx->signalError(Context::FAILURE, "Should turn on bladnn first.");
 }
 
+template <int N>
+MemRefType<int8_t, N> ral_pdll_qgemm_s8s8s8_biasadd_quant_per_tensor(
```
wyzero, line 474: maybe we can add 'nt' tag in the kernel name since it supposes that the weight is transposed before.

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
```diff
@@ -470,6 +470,65 @@ void ral_qgemm(
...
+  float beta = 1.0f;
+
+#if defined(PLATFORM_ALIBABA) and defined(ENABLE_BLADE_GEMM)
+  {
+    auto gpu_driver = ctx->getDriver<GPUDriver>(GPUDriver::name());
+    auto data =
+        static_cast<int8_t*>(gpu_driver->alloc(ctx, m * n * sizeof(int8_t)));
+    auto result = assignMemRef<int8_t, N>(data, resultSizes);
+    auto stream =
+        static_cast<se::Stream*>(gpu_driver->asSEStream(ctx, stream_handle));
```
wyzero, line 513: use `gpu_driver->asCUStream()` directly if you only want to use the underlying cuda stream handle.

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
linbinskn: done

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
linbinskn: done

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
linbinskn: Good point, done.

pytorch_blade/tests/disc/testing_base.py
```diff
@@ -211,3 +211,21 @@ def setUp(self):
...
+        cfg.enable_int8 = enable_int8
+        env_var = {}
+        if pdll_files is not None:
+            env_var["DISC_TORCH_PDL_FILES"] = pdll_files
+        if pdll_dirs is not None:
+            env_var["DISC_TORCH_PDLL_INCLUDE_DIRS"] = pdll_dirs
+        with set_env(**env_var), cfg:
+            opt_model = optimize(model, True, inp)
+        now_output = opt_model(inp)
+        self.assertTrue(torch.abs(now_output-origin_output).max() <= 0.3)
```
chenbohua3, line 231: 1. why there is a diff？
2. do not hard code the threshold
3. use `torch.allclose` instead

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
linbinskn: If inserting reshape op method has no obvious advantage compared to current implementation, I prefer to maintain the status quo since these two methods actually do the same thing and it seems no significant workload reduction by inserting op in PDL file. To achieve same target by inserting op in PDL , we have to ensure the reshape op can be eliminated in later pass and it can support the 8bit tensor datatype.

tao_compiler/mlir/xla/ral/context/stream_executor_based_impl.cc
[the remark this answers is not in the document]
minminsun: BTW, when and where will the allocated result buffer be deallocated?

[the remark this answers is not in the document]
wyzero: buffers returned from the custom call will be deallocated by RAL automatically when no one uses them.

minminsun (approved):

PR #843, opened 2022-12-06
Loop on partial writes

tanyokwok:


pytorch_blade/pytorch_blade/common_utils/tempfs.cpp
```diff
@@ -77,12 +77,24 @@ TempFile::~TempFile() {
...
 bool TempFile::WriteBytesToFile(const std::string& bytes) {
-  auto sz = ::write(fd_, bytes.data(), bytes.length());
-  if (sz != bytes.length()) {
-    LOG(ERROR) << "Failed to write content to temp file: " << GetFilename()
-               << ", error: " << strerror(errno);
-    return false;
+  ssize_t left_len = bytes.length();
+  const char* data = bytes.data();
+  errno = 0;
+  while (left_len > 0) {
```
qiuxiafei, line 83: Just ensure we're on the [same page](https://man7.org/linux/man-pages/man2/write.2.html):
In case the disc is actually full, the first `::write` trial may success with partial write, but the second trial will fail, so there won't be dead loop.

tanyokwok: image
Yes, will run until all bytes are written; otherwise, an error occurred or the disk is full.

qiuxiafei (approved):

PR #846, opened 2022-12-07
add env to control torch-blade cluster iteration

linbinskn:
Current torch_blade cluster setting(`max_iter_count=10`) has risk of early stopping before cluster final convergence. In the case of bert-base with fakequant node, the number of fusion block is 62 when `max_iter_count=10` while the number is 7 when `max_iter_count=40`. This phenomenon drastically influence the latency.
This PR aims to prevent performance drop in some cases by adding env to set specific `max_iter_count` value. Clustering policy refinement is in TODO list.

### DIFF
--- a/pytorch_blade/torch_blade/clustering/support_fusion_algorithm.py
+++ b/pytorch_blade/torch_blade/clustering/support_fusion_algorithm.py
@@ -12,7 +12,7 @@
 import torch
 from torch_blade.logging import logger
 from torch_blade.algorithm import UnionSet, NxGraph
-import os
+from torch_blade.config import Config
 
 
 class NoCycleFusedGraphBuilder(object):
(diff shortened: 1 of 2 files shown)


pytorch_blade/torch_blade/clustering/support_fusion_algorithm.py
```diff
@@ -189,7 +190,9 @@ def can_merge(u, v):
         return True
 
     graph_topolist = graph_builder.group_topolist()
-    max_iter_count = 10
+    # some graph unions may not converge in 10 iterations, provide customize setting from env
+    # TODO: refine cluster policy
+    max_iter_count = int(os.getenv('TORCH_CLUSTER_MAX_ITER_COUNT', 10))
```
qiuxiafei, line 195: I think a config option is torch_blade's convention.

qiuxiafei:
What's the initial idea of fixed iteration number instead of simply run to convergence? @tanyokwok

pytorch_blade/torch_blade/clustering/support_fusion_algorithm.py
linbinskn: Good point, done.

qiuxiafei (approved):

tanyokwok:
> What's the initial idea of fixed iteration number instead of simply run to convergence?

It's set to avoid too long and dead loops. Anyway, I think it's time to refine the subgraph capture algorithms in TorchBlade.

Issue #787, opened 2022-11-24
Introduce MLIR transform dialect to BladeDISC


wyzero:
We'll start to explore using MLIR [transform dialect ](https://mlir.llvm.org/docs/Dialects/Transform) to do codegen for compute-intensive (fused) pattern. The initial target is to support gemm codegen on ARM platform to address the dynamic shape problem of Arm Compute Library.

The initial plan is:

[ ]  Step 1,  enhance the fusion decision pass. We’ll add a new fusion kind `kTransform` for the transform-based fusion pattern.
[ ]  Step 2,  lower the lmhlo fusion op to linalg on tensor.
[ ]  Step 3,  transform the linalg computation to loops using transform dialect.
[ ]  Step 4,  refined the transformed loop to make it suitable for BladeDISC runtime.

wyzero:
e2e model test on: Bert Base (TF) and Albert (PyTorch),  on g6r, using single thread. Note that we only have one default schedule for all shape and the schedule is known to be less performant when n or k is large, thus the initial performance is supposed to be improved when we support schedule selection logic.

Bert Base (TF) 

input | TF 2.8(s) | DISC-ACL(s) | DISC-Transform(s) | speedup (DISC-transform / DISC-ACL)
-- | -- | -- | -- | --
(1, 128) | 0.742 | 0.638 | 0.656 | 97.3%
(2, 128) | 1.41 | 1.24 | 1.27 | 97.6%
(4, 128) | 2.85 | 2.36 | 2.55 | 92.5%
(8, 128) | 5.84 | 4.68 | 5.07 | 92.3%
(16, 128) | 11.9 | 9.55 | 10.2 | 93.6%
----


Albert (PyTorch)

input | TorchScript | OnnxRuntime | DISC-ACL | DISC-Transform
-- | -- | -- | -- | --
(2, 12) | 0.197 | 0.140 | 0.117 | 0.139

wyzero:
some sharing doc:

https://bladedisc.oss-cn-hangzhou.aliyuncs.com/docs/transform-dialect-based-codegen-in-bladedisc.pdf

PR #853, opened 2022-12-08
[transform] add a disc.cache_read transform primitive

wyzero:
to #787

### DIFF
--- a/tao_compiler/mlir/disc/tools/disc-transform/TransformOps/TransformOpsExt.cc
+++ b/tao_compiler/mlir/disc/tools/disc-transform/TransformOps/TransformOpsExt.cc
@@ -36,6 +36,7 @@
 #include "mlir/Dialect/Transform/IR/TransformInterfaces.h"
 #include "mlir/Interfaces/DestinationStyleOpInterface.h"
 #include "mlir/Pass/PassManager.h"
+#include "tensorflow/compiler/mlir/disc/tools/disc-transform/LinalgExt/LinalgExtOps.h"
 #include "tensorflow/compiler/mlir/disc/tools/disc-transform/utils.h"
 
 namespace mlir {
@@ -56,6 +57,7 @@ CommonExtensions::CommonExtensions() {
 using bufferization::BufferizationOptions;
 using bufferization::OneShotAnalysisState;
 using bufferization::OneShotBufferizationOptions;
+using disc_linalg_ext::MultiLevelPackOp;
 
 namespace {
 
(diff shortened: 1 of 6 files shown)


tao_compiler/mlir/disc/tools/disc-transform/TransformOps/TransformOpsExt.cc
```diff
@@ -575,6 +577,309 @@ DiagnosedSilenceableFailure FoldProducerExtractSliceOp::applyToOne(
...
+        sliceOp.getOperation(),
+        "expect the strides of target slice op are all ones\n");
+
+  // verify the target op have static shape
+  auto targetTy = targetOp->getResult(0).getType().cast<ShapedType>();
+  if (!targetTy.hasStaticShape() || targetTy.getRank() == 0)
+    return mlir::emitDefiniteFailure(
+        targetOp, "expect the targetOp has static shape with rank > 0\n");
+
+  ArrayRef<Operation*> anchorOps = state.getPayloadOps(getAnchor());
```
qiuxiafei, line 819: newbie: what is an anchor?

tao_compiler/mlir/disc/tools/disc-transform/TransformOps/TransformOpsExt.cc
```diff
@@ -575,6 +577,309 @@ DiagnosedSilenceableFailure FoldProducerExtractSliceOp::applyToOne(
...
+
+  if (!permutationVec.empty()) {
+    newOffsets =
+        disc_linalg_ext::interchange<OpFoldResult>(newOffsets, permutationVec);
+    newSizes =
+        disc_linalg_ext::interchange<OpFoldResult>(newSizes, permutationVec);
+  }
+
+  auto newShape = info.dimSizes;
+  newShape = disc_linalg_ext::interchange<int64_t>(newShape, info.permutation);
```
qiuxiafei, line 722: why only use inner most tiled-dims to do interchange to make new shape?

qiuxiafei (approved):

tao_compiler/mlir/disc/tools/disc-transform/TransformOps/TransformOpsExt.cc
wyzero: anchor op is used to decide where to place the new created `multi_level_pack` op. It's one of the inputs of cache_read op.

tao_compiler/mlir/disc/tools/disc-transform/TransformOps/TransformOpsExt.cc
wyzero: here we have already sliced a tile from the packed weight, thus only needs to transpose the tile (inner most dims) to replace the result of the original tile slice op.

PR #864, opened 2022-12-13
fix common config save/load bug

qiuxiafei:
Data type of args are converted to str after loading, leading to problematic behavior in boolean expression.


wyzero (approved):
Thanks~

PR #873, opened 2022-12-15
[transform] support fuse const weight into kTransform fusion pattern

wyzero:
to #787

### DIFF
--- a/tao_compiler/mlir/disc/transforms/fusion_utils.cc
+++ b/tao_compiler/mlir/disc/transforms/fusion_utils.cc
@@ -552,7 +552,7 @@ FusionPatternBase::FusionPatternBase(SmallVectorImpl<Operation*>& op_list)
 // fusion pattern contains.
 int FusionPatternBase::effectiveSize() {
   return llvm::count_if(
-      op_list_, [](Operation* op) { return !matchPattern(op, m_Constant()); });
+      op_list_, [](Operation* op) { return !isa<lmhlo::ConstantOp>(op); });
 }
 
 // Sorts the ops inside the fusion pattern according to the keys provided.
--- a/tao_compiler/mlir/disc/transforms/fusion_utils_transform_based.cc
+++ b/tao_compiler/mlir/disc/transforms/fusion_utils_transform_based.cc
@@ -18,10 +18,7 @@ namespace disc_ral {
 /////////// Transform based CPU FusionStrategy Implemenation ///////////
 ////////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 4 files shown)


tao_compiler/mlir/disc/transforms/fusion_utils.cc
```diff
@@ -552,7 +552,7 @@ FusionPatternBase::FusionPatternBase(SmallVectorImpl<Operation*>& op_list)
 // fusion pattern contains.
 int FusionPatternBase::effectiveSize() {
   return llvm::count_if(
-      op_list_, [](Operation* op) { return !matchPattern(op, m_Constant()); });
+      op_list_, [](Operation* op) { return !isa<lmhlo::ConstantOp>(op); });
```
qiuxiafei, line 555: why is it narrowed to `lmhlo::ConstantOp`?

tao_compiler/mlir/disc/transforms/fusion_utils_transform_based.cc
```diff
@@ -45,21 +42,49 @@ bool TransformBasedCpuFusionStrategy::isFusible(Operation* op) {
...
-  fusionPattern.setFusionType(inferredFusionType);
+
+  DenseSet<Value> dotWeights;
+  DenseSet<Operation*> supportedDotOps;
+  for (Operation* op : fusionPattern.getOpList()) {
+    // early return for the case where there are non supported ops.
+    if (!this->isFusible(op)) return true;
+    if (isSupportedDot(op)) {
+      supportedDotOps.insert(op);
+      dotWeights.insert(op->getOperand(1));
```
qiuxiafei, line 72: nit: No need to check if it's a constant op? could operand 0 be constant (seems unusual ...)?

tao_compiler/mlir/disc/transforms/fusion_utils.cc
wyzero: m_Constant only works for const on tensor level.

tao_compiler/mlir/disc/transforms/fusion_utils_transform_based.cc
wyzero: the producer of `op->getOperand(1)` is usually a alloc op, not const op. The const op is also one consumer of `op->getOperand(1)`. We this at line 79.

qiuxiafei: got it.

qiuxiafei (approved):

PR #878, opened 2022-12-16
Initial impl of pytorch quant toolkit

deeply:


tools/torch_quant/README.md
```diff
@@ -0,0 +1,61 @@
...
+## 1. Quantization Basic Concepts
+
+Deep learning model quantization is the process of reducing the precision of the weights
+and activations of a neural network model in order to reduce the memory footprint
+and computational complexity of the model. A common way to perform quantization is
+to represent the weights and activations using fewer bits than the original model,
+typically using 8 or fewer bits.
+
+A typical workflow of quantization is as following:
+- Step 0: Pre-train model
```
chenbohua3, line 14: the ptq step is not memtioned

tools/torch_quant/torch_quant/quantizer.py
```diff
@@ -0,0 +1,57 @@
...
+        self.backend = backend
+        self.tracer = tracer
+
+    def calib_gm(self,  gm: GraphModule, root: nn.Module) -> None:
+        modify_graph(gm, root, [set_qconfig, insert_act_observer])
+        toggle_observer(gm, observe=True, fake_quant=False)
+
+    def calib(self, model: nn.Module) -> nn.Module:
+        trace_mapping = fx_trace(model, self.module_filter, tracer=self.tracer)
+        for x in trace_mapping.values():
```
chenbohua3, line 34: why for-loop the `trace_mapping.values()` here

tools/torch_quant/torch_quant/graph.py
```diff
@@ -0,0 +1,234 @@
...
+        if act.op == 'call_function':
+            ob_path = f'{act.name}_ob'
+        else:
+            ob_path = f'{act.target}_ob'
+
+        ob_ctr = partial(MinMaxObserver, dtype=torch.quint8)
+        _ = ctx.get_or_create_module(ob_path, ob_ctr)
+        with ctx.gm.graph.inserting_after(act):
+            ob_node = ctx.gm.graph.call_module(ob_path, (act, ))
+        act.replace_all_uses_with(
```
chenbohua3, line 148: If enable amp, replace all uses may not be suitable. For example, given a module:
```
x  -> conv1 ->
  \  
    - -> conv2 ->
```
When only conv1 needs to be quantized, we should not change the input of conv2

tools/torch_quant/torch_quant/observer.py
```diff
@@ -0,0 +1,104 @@
...
+    def forward(self, x):
+        if self.observe:
+            min_val, max_val = torch.aminmax(x.detach().to(self.min_val.dtype))
+            self.min_val.copy_(torch.min(min_val, self.min_val))
+            self.max_val.copy_(torch.max(max_val, self.max_val))
+            scale, zero_point = self._calculate_qparams()
+            self.scale.copy_(scale)
+            self.zero_point.copy_(zero_point)
+        if self.fake_quant:
+            return torch.fake_quantize_per_tensor_affine(x, self.scale, self.zero_point, 0, 255)
```
chenbohua3, line 102: do not hard code the quant_min and quant_max

chenbohua3:
When modify the graph for calibration, is it necessary to replace the layer to be quantized with the referenced one (one that has fake-quant for the weight). PyTorch does not do it. However, in many backends (like disc, trt, neural_engine), if there is no fake-quant op corresponding to the weight on the graph, the quantization will not be enabled.

chenbohua3:
Add copyright statement at the top of each file

tools/torch_quant/torch_quant/quantizer.py
```diff
@@ -0,0 +1,57 @@
...
+            self.calib_gm(x.gm, x.m)
+        return copy_and_replace(model, trace_mapping)
+
+    def quantize_gm(self, gm: GraphModule, root: nn.Module) -> None:
+        passes = [set_qconfig, insert_act_observer,
+                  observer_to_qdq, quantizable_module_to_ref]
```
chenbohua3, line 40: Why replace observers with quantize/de-quantize? For blade-disc/blade-trt backends, replacing fake-quant with q-dq is done on torchscript.

tools/torch_quant/torch_quant/quantizer.py
deeply: The `fx_trace()` will support specifying module filters: user can explicitly include or exclude some modules in root model.  so the returned value of `fx_trace()` could be a dict like:

```
{
    'sub_module.conv_layer_0': TracePair(gm=..., m=...),
    'sub_module.conv_layer_1': TracePair(gm=..., m=...),
}
```

As for now (without filters), returned value is just `{'': TracePair(gm=..., m=root)}`

tools/torch_quant/torch_quant/quantizer.py
deeply: disc backend support has not been implemented yet in this initial commit. A simple implementation in my mind is: there will be different 'quantize passes' (or convert passes) for different backend, and common passes are those for Backend.REFERENCE. Of course it could be inappropriate or unnecessary to share common passes, I'll refactor this in next PR when supporting disc backend.

tools/torch_quant/torch_quant/graph.py
deeply: `delete_user_cb` is used to determine if the user should be switched to use new node. For example:
```
x -> conv -> 
  \ 
   - -> some_unquantizable_op ->
```
`act_nodes[x]` will only contain `conv` node, and `some_unquantizable_op` node will still connect to original `x` node after this `insert_act_observer` pass.

deeply: However, amp could be a different story. I haven't thought about how to implement amp yet  🙃
If amp is determined before this `insert_act_observer` pass, then there should be no problem. Otherwise, we may need to handle graph re-write for amp specifically.

tools/torch_quant/README.md
deeply: Updated readme, and it's more accurate to put calibration before qat.

tools/torch_quant/torch_quant/observer.py
deeply: fixed.

chenbohua3 (approved):

PR #879, opened 2022-12-16
Support transpose fusion of kDot.

JamesTheZ:
It supports single transpose in the epilogue of `kDot` fusion.

### DIFF
--- a/tao_compiler/mlir/disc/transforms/fusion_utils_dot_gpu.cc
+++ b/tao_compiler/mlir/disc/transforms/fusion_utils_dot_gpu.cc
@@ -45,6 +45,52 @@ struct DenseMapInfo<SmallVector<mlir::Operation*>> {
 namespace mlir {
 namespace disc_ral {
 
+namespace {
+
+void getDirectConsumerOps(Operation* op, DenseSet<Operation*>& consumers) {
+  consumers.clear();
+  if (op == nullptr) {
+    return;
+  }
+  SmallVector<Value> affectedValues;
+  for (auto result : op->getResults()) {
+    affectedValues.push_back(result);
+  }
+  for (auto operand : op->getOperands()) {
+    if (IsOpWriteValue(op, operand)) {
+      affectedValues.push_back(operand);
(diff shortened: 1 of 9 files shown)


tao_compiler/mlir/disc/transforms/fusion_utils_dot_gpu.cc
```diff
@@ -139,57 +184,95 @@ bool DotGpuFusionStrategy::initFusionPattern(ShapeAnalysis& shapeAnalysis,
...
-  for (Value value : op->getOperands().take_front(num_input_operand)) {
-    auto producer = fusion_pattern.findLastWriter(value);
-    if (op_set.contains(producer)) {
-      producers.insert(producer);
+    // Only support permutation of [0,2,1,3], and dim-1 is static-shaped.
+    auto perm_attr = dyn_cast<lmhlo::TransposeOp>(transpose)
+                         .getPermutation()
+                         .getValues<int64_t>();
+    SmallVector<int64_t> perm{perm_attr.begin(), perm_attr.end()};
+    if (!(perm[0] == 0 && perm[1] == 2 && perm[2] == 1 && perm[3] == 3)) {
```
qiuxiafei, line 258: nit: cutlass doesn't support other kind of transpose ?

qiuxiafei (commented):
we should have another UT to check the code is generated correctly when all conditions match.

JamesTheZ:
> we should have another UT to check the code is generated correctly when all conditions match.

The UT is in file: tao_compiler/mlir/disc/tests/regression/data/epilogue_fusion_transpose_bmm0213.mlir

tao_compiler/mlir/disc/transforms/fusion_utils_dot_gpu.cc
JamesTheZ: It supports some more but does not appear in standard BERT-like models. It can be extended in the future.

qiuxiafei (approved):

PR #880, opened 2022-12-19
enhance transpose simplifier pass

wyzero:


tao_compiler/mlir/disc/transforms/disc_transpose_simplifier.cc
```diff
@@ -882,6 +922,141 @@ LogicalResult reverseBinaryOpsIfBeneficial(Block* block, bool& changed) {
...
+      for (auto& transposedUser : transposedUsers) {
+        LLVM_DEBUG(llvm::dbgs()
+                   << "\t\t transposedUser: " << *transposedUser << "\n");
+      }
+      for (auto user : curOp->getResult(0).getUsers()) {
+        LLVM_DEBUG(llvm::dbgs() << "\tuser of curOp: " << *user << "\n");
+        if (isa<tensor::DimOp, shape::ShapeOfOp, mhlo::TransposeOp>(user))
+          continue;
+        if (transposedUsers.find(user) != transposedUsers.end()) continue;
+        needReverse = true;
```
qiuxiafei, line 989: nit: break?

tao_compiler/mlir/disc/transforms/disc_transpose_simplifier.cc
```diff
@@ -882,6 +922,141 @@ LogicalResult reverseBinaryOpsIfBeneficial(Block* block, bool& changed) {
...
+        // transpose can be folded for const op.
+        if (dyn_cast_or_null<mhlo::ConstantOp>(operandDefiningOp)) {
+          LLVM_DEBUG(llvm::dbgs()
+                     << "\t\tskip propagate for operand #" << en.index()
+                     << " due to operand op is const op\n");
+          continue;
+        }
+        bool needInsertTransposeOp =
+            tryUpdate(operandDefiningOp, curOp, operandPerm);
+        if (needInsertTransposeOp) {
```
qiuxiafei, line 1034: so the current 'cost' is number of transpose op? Not the data amount transposed ?

tao_compiler/mlir/disc/transforms/disc_transpose_simplifier.cc
wyzero: It's the last statement of the for loop body, thus I think it's not necessary to add a break here.

tao_compiler/mlir/disc/transforms/disc_transpose_simplifier.cc
qiuxiafei: Anyway, let's start here. Refine later, maybe.

qiuxiafei (approved):

PR #881, opened 2022-12-19
fix torchbench result analysis bug

Guo-Peilin:


pytorch_blade/benchmark/TorchBench/test_torch_bench.sh
```diff
@@ -85,7 +85,11 @@ do
 done
 
 # performance anaysis
-python3 results_anaysis.py -t ${results} -i ${oss_dir} -p ${RELATED_DIFF_PERCENT}
+if [ $1 == "cuda" ]; then
+  python3 results_anaysis.py -t ${results} -i ${oss_dir} -p ${RELATED_DIFF_PERCENT} -f "cuda"
+else
+  python3 results_anaysis.py -t ${results} -i ${oss_dir} -p ${RELATED_DIFF_PERCENT} -f "cpu"
```
qiuxiafei, line 91: nit: just use `-f $1` ?

pytorch_blade/benchmark/TorchBench/results_anaysis.py
```diff
@@ -13,12 +13,17 @@
...
 import pandas as pd
 
-COMPARE_FIELDS = [
+CUDA_COMPARE_FIELDS = [
     "disc (latency)",
     "blade (latency)",
     "dynamo-blade (latency)",
     "dynamo-disc (latency)",
 ]
+CPU_COMPARE_FIELDS = [
```
qiuxiafei, line 22: nit: how about just pass fields as argument of `analyze_target` ?

pytorch_blade/benchmark/TorchBench/test_torch_bench.sh
Guo-Peilin: $1 could be "cuda", "cpu" or "aarch64", so I guess we cannot just use `-f $1`

zzpmiracle (approved):
LGTM, thx！

PR #883, opened 2022-12-19
[PT2] patch diffusion model output

tanyokwok:


pytorch_blade/torch_blade/dynamo/patch_user_defined.py
```diff
@@ -0,0 +1,42 @@
...
+from torch._dynamo.eval_frame import skip_code
+
+_DataClassVariable = dicts.DataClassVariable
+
+class PatchedDataClassVariable(_DataClassVariable):
+    @staticmethod
+    @functools.lru_cache(None)
+    def _patch_once():
+        _DataClassVariable._patch_once()
+        from diffusers.models.attention import Transformer2DModelOutput
```
zzpmiracle, line 24: would this raise import error when diffusers not installed?

tanyokwok: Currently, it's ok since it will only be run after `is_matching_cls` is True. The code follows the original implementation.

zzpmiracle (approved):
LGTM!

PR #899, opened 2022-12-22
[transform] add a default schedule

wyzero:
to #787

### DIFF
--- a/tao_compiler/mlir/disc/transforms/disc_transform_legalize_to_loop.cc
+++ b/tao_compiler/mlir/disc/transforms/disc_transform_legalize_to_loop.cc
@@ -23,6 +23,7 @@ limitations under the License.
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Transform/IR/TransformInterfaces.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/Attributes.h"
 #include "mlir/IR/BlockAndValueMapping.h"
@@ -37,11 +38,14 @@ limitations under the License.
 #include "tensorflow/compiler/mlir/disc/tools/disc-transform/LinalgExt/LinalgExtDialect.h"
 #include "tensorflow/compiler/mlir/disc/tools/disc-transform/LinalgExt/LinalgExtOps.h"
 #include "tensorflow/compiler/mlir/disc/tools/disc-transform/transforms/passes.h"
+#include "tensorflow/compiler/mlir/disc/tools/disc-transform/utils.h"
 #include "tensorflow/compiler/mlir/disc/transforms/PassDetail.h"
 #include "tensorflow/compiler/mlir/disc/transforms/codegen_utils.h"
+#include "tensorflow/compiler/mlir/disc/transforms/disc_transform_schedule.h"
 #include "tensorflow/compiler/mlir/disc/transforms/fusion_utils.h"
 #include "tensorflow/compiler/mlir/disc/transforms/lhlo_elemental_utils.h"
 #include "tensorflow/compiler/mlir/disc/transforms/placement_utils.h"
+#include "tensorflow/tsl/platform/default/logging.h"
 
 // This file implements logic to legalize transform fusion pattern to loop.
 
@@ -83,8 +87,13 @@ struct DiscTransformLegalizeToLoopPass
 
   void runOnOperation() override;
 
+  void getDependentDialects(DialectRegistry& registry) const override {
+    addTransformDialectDependentDialects(registry);
+  }
+
   LogicalResult handleCpuFusionOp(OpBuilder& b, Operation* fusion,
-                                  ShapeAnalysis& shapeAnalysis);
+                                  ShapeAnalysis& shapeAnalysis,
+                                  ScheduleDispatcher& scheduleDispatcher);
   // Outlines the fusion op to a standalone module op.
   LogicalResult outlineFusionOp(lmhlo::FusionOp fusionOp,
                                 FusionPattern& fusionPattern,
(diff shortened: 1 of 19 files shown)


tao_compiler/mlir/disc/transforms/disc_transform_legalize_to_loop.cc
```diff
@@ -132,10 +141,26 @@ LogicalResult DiscTransformLegalizeToLoopPass::outlineFusionOp(
 
 LogicalResult DiscTransformLegalizeToLoopPass::runTransformPipeline(
     ModuleOp m) {
+  if (VLOG_IS_ON(1))
```
qiuxiafei, line 144: looks wired with multiple loggers together. :P

tao_compiler/mlir/disc/transforms/disc_transform_schedule.h
```diff
@@ -0,0 +1,130 @@
...
+// fusion pattern.
+class PatternDescription {
+ public:
+  explicit PatternDescription(lmhlo::FusionOp op, FusionPattern& fusionPattern,
+                              ShapeAnalysis& shapeAnalysis);
+
+  // Returns the kind of this `PatternDescription`.
+  PatternKind getPatternKind() const;
+
+  // Returns the tags attached to this fusion pattern.
```
qiuxiafei, line 49: what's the meaning of `tag`?

tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
```diff
@@ -0,0 +1,430 @@
...
+  b.setInsertionPointToStart(&bodyBlock);
+  Value variant = bodyBlock.getArgument(0);
+
+  // %fill = transform.structured.match ops{["linalg.fill"]} in %variant
+  Value fill = buildMatchOp(b, loc, variant, {"linalg.fill"});
+  // %matmul = transform.structured.match ops{["linalg.matmul"]} in %arg1
+  Value matmul = buildMatchOp(b, loc, variant, {"linalg.matmul"});
+
+  // transform.structured.tile_to_foreach_thread_op %matmul num_threads [1, 1]
+  auto forEachThreadOp = buildTileToForEachThreadOp(b, loc, matmul, {1, 1});
```
qiuxiafei, line 243: It seems to rewrite the schedule's pdl version in c++ ? Which kind do we prefer in our codebase?

tao_compiler/mlir/disc/transforms/disc_transform_legalize_to_loop.cc
wyzero: LLVM community does not support logging control in release build, which is very helpful for developers. Thus . Thus I think it's worth to introduce this feature here.

tao_compiler/mlir/disc/transforms/disc_transform_schedule.h
wyzero: see the comment line 27 in the same file.

tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
```diff
@@ -0,0 +1,430 @@
...
+  if (factoryIt == it->second.end()) return nullptr;
+  return factoryIt->second;
+}
+
+ScheduleDispatcher::ScheduleDispatcher(const std::string& transformFileName)
+    : transformFileName_(transformFileName) {}
+
+LogicalResult ScheduleDispatcher::parseModuleFromFile(MLIRContext* ctx) {
+  if (transformFileName_.empty() || !parsedModuleMap_.empty()) return success();
+  std::string expectedFormatStr =
```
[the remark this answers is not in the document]
wyzero: I think such corner case may not be very important.  Many env vars (e.g. PATH) also use special separator?

tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
wyzero: It's MLIR file with transform IR instead of PDL actually. The C++ version is used to provide a default schedule while we may override the default strategy for common fusion pattern (e.g. GEMM) by providing customized transform IR. Another use case is to provide codegen implementation for customized pattern. In that case, PDL may be useful to define the pattern.

it's worth noting that the C++ version is JIT generated. Thus it can be fine-tuned to make it suitable for specific shape setting. For example, we always do two level tiling in the current schedule. However, some tiled loops may only has 1 iteration if the shape of the GEMM is partial known (e.g., n, k of weight) and very small.  We have to take such information into consideration in order to make sure the transform IR is valid (e.g. some transform primitive depends on the number loop level).

tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
```diff
@@ -0,0 +1,430 @@
...
+
+  buildLowerVectors(b, loc, {0, 1, 2, 3, 4}, "outerproduct", "innerparallel",
+                    "linalg-copy", true, "eltwise", false);
+  buildLowerVectors(b, loc, {5, 6, 7}, "outerproduct", "innerparallel",
+                    "linalg-copy", true, "eltwise", false);
+  b.create<transform::YieldOp>(loc);
+  return success();
+}
+
+DISC_TRANSFORM_SCHEDULE(PatternKind::kGEMM, "",
```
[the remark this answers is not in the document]
wyzero: Thanks, I'll add it.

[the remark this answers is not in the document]
wyzero: I think more about this. Maybe we should just use one tag instead of adding an explicit filed for device.

For example we may have different schedule for x86 and aarch64. And even within the same cpu family, there are many different hardware features. thus using tag maybe more robust.

tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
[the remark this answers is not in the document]
qiuxiafei: In bash `:` can be used but has to be escaped. I agree it's really a corner case which may not be the highest priority.

tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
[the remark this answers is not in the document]
qiuxiafei: As discussed offline, let leave a single `tag` here, and refine it if necessary when doing scheduling selection.

qiuxiafei (approved):

PR #901, opened 2022-12-22
experimentally support compiling subgraphs in parallel

qiuxiafei:


tanyokwok:
How about adding a multi-threading stress test?

tanyokwok (approved):

PR #906, opened 2022-12-27
[transform] make the default schedule support small shape

wyzero:
to #787

### DIFF
--- a/tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
+++ b/tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
@@ -234,6 +234,38 @@ LogicalResult aarch64GEMMDefaultScheduleFactory(PatternDescription& pd,
   b.setInsertionPointToStart(&bodyBlock);
   Value variant = bodyBlock.getArgument(0);
 
+  auto& fusionPattern = pd.getFusionPattern();
+  auto dotOp =
+      dyn_cast_or_null<lmhlo::DotGeneralOp>(fusionPattern.getDominantOp());
+  if (!dotOp) {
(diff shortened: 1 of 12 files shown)


tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
```diff
@@ -272,12 +311,65 @@ LogicalResult aarch64GEMMDefaultScheduleFactory(PatternDescription& pd,
...
-  forEachThreadLoop = buildMatchOp(b, loc, variant, {"scf.foreach_thread"});
-  auto outterLoopForN = buildGetParentForOp(b, loc, padForInput, 4);
-  buildCacheRead(b, loc, padForWeight, forEachThreadLoop, {1, 1}, {1, 16}, true,
-                 {2, 0, 1, 3});
-  buildCacheRead(b, loc, padForInput, outterLoopForN, {1, 1}, {6, 1}, true,
-                 {0, 2, 3, 1});
+
+  // Check if we need to pad dimension `m/n/k` if input or weight is packed
+  bool mIsPadded = (M1 != 1) && (M == ShapedType::kDynamicSize ||
+                                 (M > M0 && (M % M0 != 0 || M0 % M1 != 0)) ||
```
qiuxiafei, line 317: In what case we'll create a schedule such that `M0 % M1 != 0` ?

qiuxiafei (approved):
LGTM with single comment~

tao_compiler/mlir/disc/transforms/disc_transform_schedule.cc
wyzero: It's just used to make the check more general. It's often not a good idea to use `M0 % M1 !=0`.

PR #912, opened 2022-12-28
[transform] not allow const as the final output of fusion

wyzero:
1, not fuse a const op if there are external consumers.

2, add a pattern to duplicate const ops for dot ops to reduce the case where multiple dot ops share the same weight op while requiring different layout.

to #787


qiuxiafei (approved):
LGTM~

PR #914, opened 2022-12-28
increase `kFoldOpEltLimit` in hlo dialect to 64GB

wyzero:
This allows to fold subgraphs with larger shape.


qiuxiafei (approved):

tanyokwok:
> This allows to fold subgraphs with larger shape.

Did it resolve any issues?

wyzero:
> > This allows to fold subgraphs with larger shape.
> 
> Did it resolve any issues?

Yes. Without this change, concat ops generated by dot merge pass will not be fold due to exceeding the limit.

PR #930, opened 2023-01-03
using lld linker

Guo-Peilin:


qiuxiafei:
Do we need to install lld in docker ?

tao/.bazelrc
```diff
@@ -37,7 +37,7 @@ build:disc --config=release_base
 
 build:disc_cpu --config=disc --cxxopt=-DTAO_CPU_ONLY
 build:disc_x86 --config=disc_cpu --config=release_cpu_linux --cxxopt=-DTAO_X86 --define disc_x86=true
-build:disc_aarch64 --config=disc_cpu --cxxopt=-DTAO_AARCH64 --define disc_aarch64=true --linkopt="-Xlinker --stub-group-size -Xlinker 10000000"
```
wyzero, line 40: Why remove these flags?

Guo-Peilin: lld do not recognize this flag

Guo-Peilin:
>

Guo-Peilin:
> Do we need to install lld in docker ?

will use default `ld.gold` if `lld` not installed

scripts/python/tao_build.py
```diff
@@ -333,6 +334,10 @@ def bazel_build(target, flag=""):
         if args.skip_compute_intensive_fusion:
             flag += ' --config=skip_compute_intensive_fusion'
 
+        if hasLLD():
```
qiuxiafei, line 337: could all `if hasLLD()`s in each file be in a common place?

qiuxiafei:
Do we need to install `lld` in our dev docker image by default?

pytorch_blade/bazel_build.py
```diff
@@ -152,6 +152,8 @@ def __init__(self, *args, **kwargs):
         if running_on_ci():
             self.configs += ["--config=ci_build"]
 
+        self.configs += [" --linkopt='-fuse-ld=gold -Xlinker --stub-group-size -Xlinker 10000000'"]
```
qiuxiafei, line 155: Nit: We could extend `extra_link_flags` to receive an default argument like `force_gold=False` or whatever, to encapsulate gold default argument into a single place.

qiuxiafei (approved):
LGTM with minor comments~

qiuxiafei (approved):

Issue #936, opened 2023-01-05
torch-blade shape anaysis related problems


zzpmiracle:
We forcely anaysis sahpe for each op, and erase shape information after failed, to get totally dynamic graph in https://github.com/alibaba/BladeDISC/pull/929 . This issue tracks realted problems occured in examples, TorchBench, and other scenes.
### shape anaysis
- [ ] https://github.com/alibaba/BladeDISC/actions/runs/3840036856/jobs/6538601043#step:6:1651
### related lowering problems
- [ ] torch.ones/zeros with dynamic size input not support lowering

tanyokwok:
ones/zeros can be decomposed into aten.full

PR #941, opened 2023-01-05
fix asr encoder forward chunk

tanyokwok:
This PR will fix the shape analysis issues of the torch.ones/zeros mentioned in https://github.com/alibaba/BladeDISC/issues/936 as well


Orion34-lanbo (approved):
LGTM

PR #943, opened 2023-01-06
Using shared memory to cache element-wise results in stitch fusion.

JamesTheZ:
This PR uses shared memory to cached intermediate data of element-wise ops that could be reused across different stitch groups. It helps to reduce redundant computation and off-chip memory traffic. It requires that `DISC_MEM_INTENSIVE_OPT_EXPERIMENTAL=true`.

### DIFF
--- a/tao_compiler/mlir/disc/transforms/lhlo_elemental_utils.cc
+++ b/tao_compiler/mlir/disc/transforms/lhlo_elemental_utils.cc
@@ -64,6 +64,24 @@ LowerConfig::SpecificLoader* LowerConfig::getSpecificLoader(Operation* op,
              : &specific_loader->second;
 }
 
+LowerConfig::SpecificStore* LowerConfig::getSpecificStore(Operation* op,
+                                                          Value operand) {
+  auto getParentFusionOp = [&](Operation* op) {
+    auto parent = op->getParentOp();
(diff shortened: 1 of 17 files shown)


tao_compiler/mlir/disc/transforms/lhlo_elemental_utils.cc
```diff
@@ -104,12 +122,20 @@ Value createLoadOrUseCachedValue(Location loc, OpBuilder* b, Operation* op,
...
                      ValueRange out_indices, LowerConfig* lower_config) {
-  if (lower_config != nullptr && lower_config->isWrittenBack(op)) {
+  if (lower_config != nullptr) {
     Value out_memref = cast<lmhlo::LmhloOp>(op).getResultBuffer();
     if (out_memref == nullptr) {
       return nullptr;
     }
-    b->create<memref::StoreOp>(loc, value, out_memref, out_indices);
+    // Write back to off-chip memory.
+    if (lower_config->isWrittenBack(op)) {
```
qiuxiafei, line 131: would also write to shared mem if `isWrittenBack(op)`?

JamesTheZ: They are different. `isWrittenBack` is to write back to global memory, not shared memory.

qiuxiafei (approved):

PR #944, opened 2023-01-06
Support decoder-only generation in hf transformers adapter.

deeply:
Refactor huggingface transformer adapter to support decoder-only generation pipeline (e.g. text-generation with gpt2):
- programmatically create wrapper class (inherited from original hf transfromers model class) to keep generation related methods (e.g. prepare_inputs_for_generation).
- specify default forward kwargs (e.g. use_cache) so we can torch.jit.trace model correctly.
- add some unit tests.

Some TODOs (will be implemented in future PRs):
- support use_cache=True for decoder-only model
- support encoder-decoder generation model (e.g. bart, t5)

### DIFF
--- a/examples/PyTorch/Inference/hf_transformers/blade_adapter.py
+++ b/examples/PyTorch/Inference/hf_transformers/blade_adapter.py
@@ -16,6 +16,7 @@ from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Type, Union
 
 import torch
 import torch_blade
+from makefun import create_function
 from transformers import (AutoConfig, AutoFeatureExtractor, AutoTokenizer,
                           FeatureExtractionMixin, Pipeline, PretrainedConfig,
                           PreTrainedModel, PreTrainedTokenizer, ProcessorMixin)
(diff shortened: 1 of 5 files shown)


examples/PyTorch/Inference/hf_transformers/tests/test_wrapper_model.py
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+from blade_adapter import WrapperModel, create_model, trace_model
+from transformers import GPT2LMHeadModel
+
+
+class WrapperModelTest(unittest.TestCase):
```
tanyokwok, line 18: Where is the script that triggers those tests?

deeply: These tests haven't been added to CI yet. Considering some tests need to download models(typically hundreds of MBs per model) and run disc compilation, and code change in this adapter won't effect disc itself, we may want to setup a separated test schedule for this?

examples/PyTorch/Inference/hf_transformers/blade_adapter.py
```diff
@@ -197,6 +198,16 @@ def forward(self, *args, **kwargs) -> Any:
         outputs = self.opt_model(*model_args)
         return ModelOutput({k: v for k, v in zip(self.info.output_names, outputs)})
 
+    @staticmethod
+    def create_wrapper_model(opt_model: torch.jit.ScriptModule, info: ModelInfo) -> 'WrapperModel':
```
qiuxiafei, line 202: fancy ...

qiuxiafei (approved):

PR #958, opened 2023-01-12
Support DISC backend in torch_quant tool

deeply:


tools/torch_quant/torch_quant/observed_module.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+Linear = type('Linear', (nnqat.Linear, ), {
+              'from_float': classmethod(_from_float), 'forward': _linear_forward})
+
+OB_MODULE_MAPPING = {
+    nn.Conv1d: Conv1d,
+    nn.Conv2d: Conv2d,
+    nn.Conv3d: Conv3d,
+    nn.Linear: Linear,
```
linbinskn, line 53: Is it feasible to add func module like `torch.matmul` in current design?

tools/torch_quant/bert_ptq_demo.py
```diff
@@ -41,27 +47,90 @@ def calibrate(model, dataloader, device):
...
+    print('load model...')
+    model = AutoModelForSequenceClassification.from_pretrained(
+        MODEL_ID, torchscript=True)
+    # model.config.problem_type is set in first forward calling, and is required for fx tracing
+    # we call calibrate() here only as a dummy forward calling.
+    calibrate(model, calib_dataloader, device)
+
+    tracer = HFTracerWrapper(
+        ["input_ids", "attention_mask", "token_type_ids"])
+    quantizer = Quantizer(tracer=tracer, backend=Backend.DISC)
```
linbinskn, line 107: Is there any interface for us to set the quant module list?

tools/torch_quant/torch_quant/observed_module.py
deeply: I think supporting quantizable functions like torch.matmul, is similar with supporting quantizable modules like nn.Linear. Need to implement it with a new "graph modification pass" though. Similar logic like `quantizable_module_to_observed` : find the call_function node in fx graph, and insert observer after weight and bias.

tools/torch_quant/bert_ptq_demo.py
deeply: There is a `module_filter` argument for `Quantizer` for this purpose. Not implemented yet 😂

tools/torch_quant/torch_quant/graph.py
```diff
@@ -162,6 +165,34 @@ def insert_act_observer(ctx: GraphModContext) -> None:
     ctx.gm.recompile()
 
 
+def quantizable_module_to_observed(ctx: GraphModContext) -> None:
+    """
+    Replace quantizable modules with observed version.
+
+    For quantizable modules (e.g. Conv2D, Linear), weight is stored and used
+    internally, instead of as a sperated node in fx graph. So we need to
```
chenbohua3, line 173: typo: `sperated`

tools/torch_quant/torch_quant/observed_module.py
```diff
@@ -0,0 +1,54 @@
...
+    new_mod.bias_ob = bias_ob
+    delattr(new_mod, 'weight_fake_quant')
+    return new_mod
+
+
+def _conv_type(name: str, base: Type) -> Type:
+    def forward(self, input):
+        return self._conv_forward(input, self.w_ob(self.weight), self.bias_ob(self.bias))
+
+    return type(name, (base,), {'from_float': classmethod(_from_float), 'forward': forward})
```
chenbohua3, line 34: fancy:)

tools/torch_quant/torch_quant/quantizer.py
```diff
@@ -27,16 +31,43 @@ class Backend(Enum):
     FBGEMM = 2
 
 
+DEFAULT_ACT_OB_CTR: Dict[Backend, Callable[..., Observer]] = {
```
chenbohua3, line 34: 1. `qscheme` is not set which will lead the `symmetric` of `MinMaxObserver` always `False` and further leads the wrong scales and zero_points.
2. Not a change introduced by this pull request but will affect accuracy, [here](https://github.com/alibaba/BladeDISC/blob/6c270d03d2521acb3a999f7d9dc8ad55efacfffb/tools/torch_quant/torch_quant/observer.py#L93), the zero point should to int data type.

tools/torch_quant/torch_quant/observer.py
```diff
@@ -73,7 +74,7 @@ def __init__(self, dtype: torch.dtype = torch.qint8,
         self.register_buffer("min_val", torch.tensor(float("inf")))
         self.register_buffer("max_val", torch.tensor(float("-inf")))
         self.register_buffer("scale", torch.tensor(1.))
-        self.register_buffer("zero_point", torch.tensor(0.))
+        self.register_buffer("zero_point", torch.tensor(0, dtype=torch.int32))
```
chenbohua3, line 77: Should we support lower versions of torch. In earlier versions of torch, the zp is of `torch.long` dtype

tools/torch_quant/torch_quant/quantizer.py
chenbohua3: And better to add some uts to check the scale & zero_point after the calibration step

tools/torch_quant/bert_ptq_demo.py
```diff
@@ -22,6 +26,8 @@
 
 logging.basicConfig(level=logging.DEBUG)
 
+MODEL_ID = 'bert-base-cased-finetuned-mrpc'
```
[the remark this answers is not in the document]
deeply: It may take some time to evaluate the accuracy of demo model. I'll leave a TODO here to use a smaller model as demo.

tools/torch_quant/torch_quant/graph.py
deeply: fixed

tools/torch_quant/torch_quant/quantizer.py
deeply: > 

1. OK, qscheme is specified for each backend observer constructors in the updated commit.
2. zero point will be implicitly rounded when we copy it to a tensor with dtype=int32?

deeply: there is a very basic test at [test_min_max_observer.py](https://github.com/alibaba/BladeDISC/blob/6c270d03d2521acb3a999f7d9dc8ad55efacfffb/tools/torch_quant/tests/test_min_max_observer.py#L19). I'll add a todo for more tests.

tools/torch_quant/torch_quant/observer.py
deeply: I can't make a decision right now. maybe we can discuss how to support old versions (and which versions) when some use cases come out 😂

chenbohua3:
LGTM

chenbohua3 (approved):

PR #967, opened 2023-01-17
[transform] convert foreach thread op to parallel right before outline cpu kernel

wyzero:
single iteration `scf.parallel` op will be removed, making the outline cpu kernel pass failed to outline transform based codegen kernel (we only support single thread a.t.m.). This PR convert foreach thread op to parallel op right before output cpu kernel pass to avoid it's being removed.

to #787


qiuxiafei (approved):
LGTM~~

PR #978, opened 2023-01-31
[PT2.0] Support paraformer

tanyokwok:
The PR is needed to compile paraformer:

1. add support to aten::pad
2. inference `aten.split.Tensor` output's chunk number for the static shape of the input


Yancey1989 (approved):
LGTM

PR #980, opened 2023-01-31
separate preparing & computing process of calling onednn kernel

chenbohua3:
So we can get the primitive states of one-dnn and further make it more easier for enabling weight pre-packing


wyzero (approved):
LGTM

PR #982, opened 2023-01-31
Add DISC mhlo cse pass

tanyokwok:
The original one provided in LLVM will not work on operations with body regions such as mhlo::ReduceOp. The PR adds CSE for mhlo::ReduceOp and creates the interface to add more other mhlo operations in the future.


tao_compiler/mlir/disc/transforms/disc_cse.cc
```diff
@@ -0,0 +1,213 @@
...
+        lhs, rhs,
+        /*mapOperands=*/OperationEquivalence::exactValueMatch,
+        /*mapResults=*/OperationEquivalence::ignoreValueEquivalence,
+        OperationEquivalence::IgnoreLocations);
+  }
+};
+}  // namespace
+
+namespace {
+struct DiscMhloCSEPass : public DiscMhloCSEPassBase<DiscMhloCSEPass> {
```
Yancey1989, line 143: can you provide more messages that why we need disc-cse  pass, instead of llvm-ces?

tanyokwok: The original one provided in LLVM will not work on operations with body regions such as `mhlo::ReduceOp`. The PR adds CSE for `mhlo::ReduceOp` and creates the interface to add more other mhlo operations in the future.

tao_compiler/mlir/disc/transforms/disc_cse.cc
```diff
@@ -0,0 +1,219 @@
...
+                               rhsBody.front().getOperations())) {
+        auto& lhsOp = std::get<0>(it);
+        auto& rhsOp = std::get<1>(it);
+        // We currently do not compare equality of nested regions
+        if (lhsOp.getNumRegions() != 0 or rhsOp.getNumRegions() != 0)
+          return false;
+
+        auto innerOpMatch = OperationEquivalence::isEquivalentTo(
+            &lhsOp, &rhsOp,
+            /*mapOperands=*/OperationEquivalence::ignoreValueEquivalence,
```
wyzero, line 117: I think it's not safe to ignore operand value check here.

For example:
```

mhlo.reduce(...) {
^bb0(%arg0 : f32, %arg1 : f32) {
   %0 = arith.mulf %arg0, %c2 // const is 2
   %1 = arith.muf %0, %arg1
   return %1 : f32 
 }

mhlo.reduce(...) {
^bb0(%arg0 : f32, %arg1 : f32) {
   %0 = arith.mulf %arg0, %c3 // here is const 3
   %1 = arith.muf %0, %arg1
   return %1 : f32 
 }
```

maybe we can support more restricted reduce op, for example those only have one known op (e.g. add/sub/mul/div) in the body.

tao_compiler/mlir/disc/disc_compiler.cc
```diff
@@ -376,12 +376,21 @@ LogicalResult LowerHLOToLLVM(ModuleOp m, const DISCLoweringOptions& options) {
...
   }
-  // Run CSE after conv rewriter pass to eliminate some redundant transpose ops.
+  // Run CSE after rewriter pass to eliminate some redundant ops.
   pm.addNestedPass<FuncOp>(createCanonicalizerPass());
+  pm.addNestedPass<FuncOp>(disc_ral::createDiscMhloCSEPass());
   pm.addNestedPass<FuncOp>(createCSEPass());
   pm.addNestedPass<FuncOp>(createCanonicalizerPass());
   pm.addNestedPass<FuncOp>(disc_ral::createTransposeSimplifierPass());
 
+  if (gpu_enabled) {
```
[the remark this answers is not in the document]
tanyokwok: Removed

tao_compiler/mlir/disc/transforms/disc_cse.cc
```diff
@@ -0,0 +1,219 @@
...
+      return false;
+
+    auto lhsReduceOp = dyn_cast<mhlo::ReduceOp>(lhs);
+    auto rhsReduceOp = dyn_cast<mhlo::ReduceOp>(rhs);
+    if (lhsReduceOp && rhsReduceOp) {
+      auto& lhsBody = lhsReduceOp.getBody();
+      auto& rhsBody = rhsReduceOp.getBody();
+      if (not(lhsBody.hasOneBlock() && rhsBody.hasOneBlock())) return false;
+
+      for (auto it : llvm::zip(lhsBody.front().getOperations(),
```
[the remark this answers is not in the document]
tanyokwok: Add op number check

tao_compiler/mlir/disc/transforms/disc_cse.cc
tanyokwok: Indeed, `OperationEquivalence::isEquivalentTo` will compare op properties first. Thus different Constant values will not be regarded as equivalent. See https://mlir.llvm.org/doxygen/OperationSupport_8cpp_source.html#L712

Also, I added a test to the case you mentioned above, see `tests/disc_mlir/disc_cse.mlr:multiple_reduce_with_cst0`

wyzero: constant op may not defined inside the region of reduce op (in my original example above). Is it still ok in such case?

```
    %3 = mhlo.constant dense<1.000000e+00> : tensor<f16>
    %4 = mhlo.reduce(%arg0 init: %0) across dimensions = [0, 2, 3] : (tensor<?x?x?x?xf16>, tensor<f16>) -> tensor<?xf16>
     reducer(%arg1: tensor<f16>, %arg2: tensor<f16>)  {
      %6 = mhlo.multiply %arg1, %3 : tensor<f16>
      %7 = mhlo.add %6, %arg2 : tensor<f16>
      mhlo.return %7 : tensor<f16>
    }
```

wyzero (approved):

PR #994, opened 2023-02-06
Add config option to force multiple cuda devices into a single one

qiuxiafei:
In some case, a model may be exported with multiple devices. While for inference, only a single device is required. This option is to force all constant on all CUDA devices to the specified one.
It is potentially unsafe, just like `enable_force_to_cuda`, and should be used with care.


tanyokwok:
Could you give some background of the PR?

qiuxiafei:
> Could you give some background about the PR?

update.

tanyokwok (approved):
LGTM

PR #996, opened 2023-02-06
add clone const pass for fakequant op

Guo-Peilin:


tao/tao_bridge/passes/tao_clone_constants_for_better_clustering.cc
```diff
@@ -93,6 +93,27 @@ StatusOr<bool> IsSmallHostConstant(Node* n) {
...
+  }
+  DeviceNameUtils::ParsedName parsed;
+  TF_RET_CHECK(
+      DeviceNameUtils::ParseFullName(n->assigned_device_name(), &parsed));
+  if (parsed.type != DEVICE_CPU) {
+    return false;
+  }
+
+  std::string nodeName = n->name();
+  if (nodeName.find("fake_quantize") != std::string::npos &&
```
qiuxiafei, line 110: Is this robust enough? Only check by node name?

qiuxiafei (commented):
Why this is not done in MHLO level? We need to do it another time for pytorch?

tao/tao_bridge/passes/tao_clone_constants_for_better_clustering.cc
Guo-Peilin: this is a prototype patch right now, will add more checks later.

wyzero:
> Why this is not done in MHLO level? We need to do it another time for pytorch?

We need to clone these const before clustering. it's too later to clone const in mhlo level.

tao/tao_bridge/passes/tao_optimization_pass.cc
```diff
@@ -128,6 +128,12 @@ Status TaoOptimizationPass::Run(const GraphOptimizationPassOptions& options) {
     TF_RETURN_IF_ERROR(defuse_pass.Run(options));
   }
 
+  DumpGraph(options, "before_tao_clone_pass");
+  // Add Clone Constants Pass for fakequant
+  TaoCloneConstantsForBetterClusteringPass clone_pass(false, true);
```
wyzero, line 133: maybe it's time to remove these tvm related code?  @qiuxiafei @minminsun

qiuxiafei: I'm working on it.

tao/tao_bridge/passes/tao_clone_constants_for_better_clustering.cc
```diff
@@ -93,6 +93,26 @@ StatusOr<bool> IsSmallHostConstant(Node* n) {
...
+  if (!n->IsConstant()) return false;
+
+  DeviceNameUtils::ParsedName parsed;
+  TF_RET_CHECK(
+      DeviceNameUtils::ParseFullName(n->assigned_device_name(), &parsed));
+  if (parsed.type != DEVICE_CPU) return false;
+
+  // multi DiscFakeQuant use one Const
+  if (n->type_string() == "Const" &&
+      std::all_of(n->out_nodes().begin(), n->out_nodes().end(), [](Node* out) {
```
wyzero, line 108: It's not necessary to require all the outputs to be DiscFakeQuant. maybe just check there are DiscFakeQuant user and the number of consumers of const is larger than 1?

Guo-Peilin: fixed

wyzero (approved):

PR #997, opened 2023-02-07
[transform] epilogue fusion for gemm schedule w/o split k

wyzero:
to #787


qiuxiafei (approved):
Generally LGTM, except some similar code to other ScheduleFactories. May be some refactor in the future.

PR #1005, opened 2023-02-09
[torch-quant] add per-channel min/max observer

chenbohua3:


tools/torch_quant/torch_quant/observer.py
```diff
@@ -116,6 +170,60 @@ def forward(self, x):
...
+        self.register_buffer("max_val", torch.tensor([]))
+        self.register_buffer("scale", torch.tensor([]))
+        self.register_buffer("zero_point", torch.tensor([], dtype=torch.int32))
+        self.ch_axis = ch_axis
+
+    @property
+    def qparams(self) -> QParams:
+        return QParams(qscheme=self.qscheme, dtype=self.dtype,
+                       scale=self.scale,
+                       zero_point=self.zero_point)
```
deeply, line 187: move scale, zero_point, and qparams() to base Observer class?

tools/torch_quant/torch_quant/observer.py
```diff
@@ -116,6 +170,60 @@ def forward(self, x):
             return x
 
 
+class PerChannelMinMaxObserver(Observer):
+    def __init__(self, ch_axis=0, dtype: torch.dtype = torch.qint8,
```
deeply, line 174: Is 0 a reasonable default value for ch_axis? make ch_axis a required argument instead?

tools/torch_quant/torch_quant/observer.py
```diff
@@ -116,6 +170,60 @@ def forward(self, x):
...
+    def forward(self, x):
+        if self.observe:
+            if x.numel() == 0:
+                return x
+            x = x.detach()  # avoid keeping autograd tape
+            min_val = self.min_val
+            max_val = self.max_val
+            x_dim = x.size()
+
+            new_axis_list = [i for i in range(len(x_dim))]  # noqa: C416
```
deeply, line 198: nit: list(range(...))

tools/torch_quant/torch_quant/observer.py
chenbohua3: done

tools/torch_quant/torch_quant/observer.py
chenbohua3: Yes it is a reasonable default value because the first dim of the weight of most quantizable layers in PyTorch is corresponding to  the output channel dimention. And 0 is also adopted as a default value in the PyTorch's per-channel observer/fake-quantizer.

tools/torch_quant/torch_quant/observer.py
chenbohua3: done

deeply:
LGTM

deeply (approved):

PR #1007, opened 2023-02-09
[transform] add disc_linalg_ext.conditional_generic op

wyzero:
conditional_generic op is used to model following semantic.

```
  %new_outs = disc_linalg_ext.conditional_generic(%pred, %ins, %outs)

  // ----

  %merged_outs = if (%pred) {
    %new_outs = linalg.generic_op(%ins, %outs)
    return new_outs
  } {
    return outs
  }
```

Note that we do not use `scf.if` or `arith.select` directly due to following reason.

- We do not want to execute the computation captured by the generic op, thus we can not use `arith.select`

- We want to furture transform (e.g. tile, pad, vectorize) the computation. `scf.if` is a general control flow op and not designed to do such transform again.

The implementation of conditional_generic is very similar to linalg.generic op and thus most related code is just copied from llvm repo. We may choose to upstream such op in the future to reduce maintain overhead.

We introduce such op to address output fusion generic ops into a reduction loop. An example is shonw as following.

original code:

```
// reduction loop,
%0 = scf.for %iv0 = %c0 to %d0 step %c288 iter_args(%out = %arg0) -> (tensor<?x?xf32>) {
  %ins = tensor.exectract ...
  %new_out = linalg.matmul(%ins, %out)
  scf.yield %new_out : tensor<?x?xf32>
}
%1 = linalg.generic(..., %0, ...) {
  // some element-wise epilogue computations
}
```

after epilogue fusion:
```
// reduction loop,
%0, %1 = scf.for %iv0 = %c0 to %d0 step %c288 iter_args(%out = %arg0, %out1 = ...) -> (tensor<?x?xf32>) {
  %ins = tensor.exectract ...
  %new_out = linalg.matmul(%ins, %out)
  %pred = ... // is last iteration
  %maybe_updated_new_out = conditional_generic(%pred, ..., %0, ....) {
    // some element-wise epilogue computations
  }
  scf.yield %new_out, %maybe_updated_new_out : tensor<?x?xf32>, tensor<?x?xf32>
}
```

### DIFF
--- a/tao_compiler/mlir/disc/tools/disc-opt/disc-opt.cc
+++ b/tao_compiler/mlir/disc/tools/disc-opt/disc-opt.cc
@@ -33,6 +33,7 @@ limitations under the License.
 #include "mlir/disc/IR/hlo_disc_ops.h"
 #include "mlir/disc/IR/lhlo_disc_ops.h"
 #include "mlir/disc/tools/disc-transform/LinalgExt/LinalgExtDialect.h"
+#include "mlir/disc/tools/disc-transform/LinalgExt/LinalgExtOps.h"
 #include "mlir/disc/tools/disc-transform/TransformOps/TransformOpsExt.h"
 #include "mlir/disc/tools/disc-transform/transforms/register_passes.h"
 #include "mlir/disc/transforms/register_passes.h"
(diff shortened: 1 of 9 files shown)


qiuxiafei (approved):
LGTM~

PR #1013, opened 2023-02-10
[WIP] Initial impl of HF Diffusers adapter

deeply:


bonlime:
What's the status of this MR? I've tried using, but was unable to trace/script the components of StableDiffusion

PR #1021, opened 2023-02-16
fix torch_quant ci

chenbohua3:


tools/torch_quant/torch_quant/observer.py
```diff
@@ -214,24 +251,30 @@ def forward(self, x):
 
 
 class BiasObserver(Observer):
-    def __init__(self, w_ob: Observer, act_ob: Observer,
-                 dtype: torch.dtype = torch.qint32,
-                 qscheme: torch.qscheme = torch.per_tensor_affine, **kwargs) -> None:
-        super().__init__(**kwargs)
+    def __init__(self, w_ob: Observer, act_ob: Observer, **kwargs) -> None:
+        dtype = torch.qint32
+        is_w_per_channel = is_per_channel(w_ob.qscheme)
```
deeply, line 256: just use `w_ob.per_channel`?

tools/torch_quant/torch_quant/observer.py
```diff
@@ -66,10 +98,16 @@ def __init__(self, dtype, qscheme, **kwargs) -> None:
         self.register_buffer('eps', torch.tensor([torch.finfo(torch.float32).eps]))
         self.observe = True
         self.fake_quant = True
+        self.ch_axis = ch_axis
+        self._register_load_state_dict_pre_hook(partial(pre_load_state_dict_hook, self))
 
     @property
     def symmetric(self) -> bool:
-        return self.qscheme in (torch.per_tensor_symmetric, torch.per_channel_symmetric)
+        return is_per_channel(self.qscheme)
```
deeply: symmetric doesn't always equal to per_channel?

deeply: never mind, I see the new commit fix this.

tools/torch_quant/torch_quant/observer.py
chenbohua3: Generally, only the qscheme of weight will be per-channel.
Have added activation qscheme to co-determine the qscheme of bias

chenbohua3:
A todo has been added to remind us to find a more efficient way to deal with the shape of the scale

tools/torch_quant/torch_quant/observer.py
deeply: ok, but code can be simplified as `if w_ob.per_channel or act_ob.per_channel:` ?

chenbohua3: done

deeply (approved):
LGTM

PR #1034, opened 2023-02-21
fix bazel OOM problems

Guo-Peilin:


tensorflow_blade/.bazelrc
```diff
@@ -69,6 +69,8 @@ build:disc_aarch64 --cxxopt=-DTAO_AARCH64
 build:disc_aarch64 --define disc_aarch64=true
 build:disc_aarch64 --linkopt="-Xlinker --stub-group-size -Xlinker 10000000"
 build:disc_aarch64 --action_env BUILD_WITH_AARCH64=1
+build:disc_aarch64 --experimental_local_memory_estimate
+build:disc_aarch64 --jobs=10
```
wyzero, line 73: is `10` too small for parallelism? @qiuxiafei @Yancey1989

chenbohua3: what about determining this number according to the number of the cores on the machine

qiuxiafei: Yes, 10 is usually to small. And as on different machines, number of cores varies. How about `--local_cpu_resources`/`--local_ram_resources` here: https://bazel.build/docs/user-manual#local-resources

PR #1061, opened 2023-03-07
[torch-quant] update weight observer qscheme before converting to ref

xiaowan0322:
[torch-quant] update weight observer qscheme before converting to ref, change symmetric to affine

### DIFF
--- a/tools/torch_quant/torch_quant/graph.py
+++ b/tools/torch_quant/torch_quant/graph.py
@@ -247,7 +247,16 @@ def quantizable_module_to_ref(ctx: GraphModContext) -> None:
                 f'module type {type(src)} is not supported for static quantization.')
         w_ob = src.qconfig.weight()
         w_ob(src.weight)
-        dst = dst_type.from_float(src, w_ob.qparams._asdict())
+        # update qscheme, since we don't have symmetric quant qscheme in quantized Tensor
+        # https://github.com/pytorch/pytorch/blob/v1.13.1/torch/ao/quantization/utils.py#L138
+        wq_dict = w_ob.qparams._asdict()
+        sym_to_aff_map = {
+            torch.per_tensor_symmetric: torch.per_tensor_affine,
+            torch.per_channel_symmetric: torch.per_channel_affine,
+        }
+        wq_dict['qscheme'] = sym_to_aff_map.get(wq_dict['qscheme'], wq_dict['qscheme'])
+        wq_dict['axis'] = wq_dict.pop('ch_axis')
+        dst = dst_type.from_float(src, wq_dict)
         # TODO(litan.ls): copy forward hooks
         ctx.replace_module(node.target, dst)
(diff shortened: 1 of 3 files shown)


tools/torch_quant/torch_quant/graph.py
```diff
@@ -247,7 +247,16 @@ def quantizable_module_to_ref(ctx: GraphModContext) -> None:
...
         w_ob(src.weight)
-        dst = dst_type.from_float(src, w_ob.qparams._asdict())
+        # update qscheme, since we don't have symmetric quant qscheme in quantized Tensor
+        # https://github.com/pytorch/pytorch/blob/v1.13.1/torch/ao/quantization/utils.py#L138
+        wq_dict = w_ob.qparams._asdict()
+        sym_to_aff_map = {
+            torch.per_tensor_symmetric: torch.per_tensor_affine,
+            torch.per_channel_symmetric: torch.per_channel_affine,
+        }
+        wq_dict['qscheme'] = sym_to_aff_map.get(wq_dict['qscheme'], wq_dict['qscheme'])
```
chenbohua3, line 257: The qscheme of weight is already `affine`, why should we add some logic to ensure this here.
And will there be any problems that use `symmetric` in calibration and change it to `affine` when convert the module to the ref one

tools/torch_quant/torch_quant/quantizer.py
```diff
@@ -50,9 +50,6 @@ class Backend(Enum):
 }
 
 DEFAULT_W_OB_CTR = {
-    # According to the url below, PyTorch's reference module does not support
```
chenbohua3, line 53: Why delete this?

tools/torch_quant/torch_quant/graph.py
xiaowan0322: If backend is fbgemm or disc, the default weight qscheme is `per_channel_symmetric`. But symmetric quantized Tensor is not supported, so we should convert `symmetric` to `affine` here. Have we done this conversion before?
Generally, `symmetric` can be seen as a special case of `affine`. And the scales and zero points here are already certain values, the final result cannot be effected.

tools/torch_quant/torch_quant/quantizer.py
xiaowan0322: I thought this code modifications could explain the confusing question, but didn't it?

tools/torch_quant/torch_quant/graph.py
chenbohua3: If use DISC backend, we will use `torch.fake_quant_xxx`  instead of the really quantized tensor, so the `per_channel_symmetric ` is supported

xiaowan0322: All right, while for DISC backend, we won't convert quantizable modules to ref version.

chenbohua3 (approved):

PR #1078, opened 2023-03-15
fix unsigned type lowering

wyzero:


scripts/python/tao_build.py
```diff
@@ -409,13 +409,14 @@ def bazel_test(target, flag=""):
                 flag += ' --config=disc_mkldnn'
             if args.platform_alibaba:
                 flag += " --config=platform_alibaba"
+            flag += ' --test_tag_filters=-gpu'
```
qiuxiafei, line 412: So we been executing gpu tests on CPU ...

qiuxiafei (approved):
LGTM

scripts/python/tao_build.py
wyzero: tests with gpu-only-*  prefix will be not be executed on cpu before. maybe the behavior has been changed during rebasing or de-coupling.

PR #1088, opened 2023-03-21
[torch-quant]fix bias fake-quant missing when quantize model

chenbohua3:


tools/torch_quant/torch_quant/graph.py
```diff
@@ -283,10 +292,13 @@ def quantizable_module_to_observed(ctx: GraphModContext) -> None:
         w_ob_path = f'{node.target}.w_ob'
         w_ob = ctx.get_or_create_module(w_ob_path, ctx.w_ob_ctr)
         bias_ob = None
-        if getattr(src, 'bias', None) is not None and ctx.bias_ob_ctr:
-            bias_ob_ctr = partial(ctx.bias_ob_ctr, w_ob, act_ob)
+        if getattr(src, 'bias', None) is not None and is_observe_bias:
+            # If we use the existed bias observer, ctx.bias_ob_ctr should be None
             bias_ob_path = f'{node.target}.bias_ob'
-            bias_ob = ctx.get_or_create_module(bias_ob_path, bias_ob_ctr)
+            bias_ob = ctx.get_existed_module(bias_ob_path)
```
xiaowan0322, line 298: nitpicking: could be simplified to `bias_ob = ctx.modules.get(bias_ob_path)` ?

tools/torch_quant/torch_quant/quantizer.py
```diff
@@ -264,7 +264,7 @@ def qat_gm(
             # Generally we do not add fake-quant to bias during qat fine-tuning. If
             # users want to evaluate the accuracy of a model in specific state, they
             # should use the model returned by `quantizer.quantize`
-            quantizable_module_to_observed])
+            partial(quantizable_module_to_observed, is_observe_bias=False)])
```
xiaowan0322, line 267: For qat, `is_observe_bias` is always `False`, `bias_ob_ctr` is no longer needed for `qat()`?

tools/torch_quant/torch_quant/graph.py
chenbohua3: done

tools/torch_quant/torch_quant/quantizer.py
chenbohua3: For now, adding fake-quant for bias will slow down the training process and the benefits of doing so is unknown. So the fake-quant is removed. If any benefits is demonstrated, we can enable this.

xiaowan0322 (approved):
LGTM

PR #1091, opened 2023-03-22
[torch-quant] calculate qparams only once

xiaowan0322:
Calculate qparams only once for HistogramObserver, which could prove a great time-saver. For asr model, it could be finished in about one fourth of the time.


deeply:
We can implement this as a lazy qparam property in Observer(base class)?  and then we don't need to check qparam calculated everywhere.

xiaowan0322:
> We can implement this as a lazy qparam property in Observer(base class)? and then we don't need to check qparam calculated everywhere.

Sorry for my late reply. I updated the commit, but didn’t implement it as a lazy property. Because I think it maybe not unsuitable for 'qparam' to be a simple property in Observer.

deeply (approved):

PR #1093, opened 2023-03-22
[torch-quant] fix qat for reference and fbgemm backends

xiaowan0322:


tools/torch_quant/tests/test_bias_releated_situations.py
```diff
@@ -23,7 +23,6 @@ def test_bias_exist(self):
         dummy = torch.randn(1, 3)
         quantizer = Quantizer(backend=Backend.DISC)
         calib_model = quantizer.calib(model)
-        self.assertTrue(isinstance(model.linear.bias_ob, BiasObserver))
```
chenbohua3, line 26: For disc backend, `bias_ob` should be checked in this ut. Add a standalone or expand ut about the fbgemm or reference backends.

tools/torch_quant/torch_quant/quantizer.py
```diff
@@ -155,68 +138,71 @@ def get_observer_types(
...
-                 device: Device = Device.X86,
-                 tracer: Optional[Tracer] = None,
-                 ) -> None:
+    def __init__(
+        self,
+        module_filter: Optional[ModuleFilter] = None,
+        backend: Backend = Backend.REFERENCE,
+        device: Device = Device.X86,
+        tracer: Optional[Tracer] = None,
+        act_ob_ctr: Optional[Callable[..., Observer]] = None,
```
chenbohua3, line 147: In the current design, the kinds of observer types can be changed in each call of `qat`, `ptq`, `convert`. If put the kinds of observer here, how do we switch observer types.

tools/torch_quant/tests/test_bias_releated_situations.py
xiaowan0322: For disc backend, `bias_ob` is also not inserted in `calib` stage, which is updated in this pr.

tools/torch_quant/torch_quant/quantizer.py
xiaowan0322: Generally, the observer types in `quantize` stage should be the same as previous stage (calib, qat...). As for qat, observer types can be also changed by default or specifying.

tools/torch_quant/torch_quant/quantizer.py
```diff
@@ -155,68 +138,71 @@ def get_observer_types(
...
+        w_ob_ctr: Optional[Callable[..., Observer]] = None,
+        bias_ob_ctr: Optional[Callable[..., Observer]] = None,
+        qat_ob_ctr: Optional[Callable[..., Observer]] = None,
+    ) -> None:
+        if backend == Backend.FBGEMM and torch.backends.quantized.engine != 'fbgemm':
```
chenbohua3, line 152: If fbgemm is not available on the training platform  and the users want to deploy the trained model on a fbgemm-available platform, will this limitation too strict？

tools/torch_quant/torch_quant/quantizer.py
chenbohua3: I am afraid not. For example, recently we are doing some quantization experiments. In these experiments, the PTQ algorithm based on mathematical statistics (e.g. gptq)will be used to initialize the quantitative parameters of the model. Then, some tuning based PTQ algorithms like brecq will be used to further improve the performance. If doing this experiments on torch-quant, the workflow should be:
```
model = some_llm_models()

gptq_quant_model = quantizer.calib(model)

do calibration

brecq_quant_model = quantizer.calib(model, w_ob_ctr=brecq_ctr)

do calibration
```

So the interface to set the ob_ctr should be available on both ptq/qat function.

tools/torch_quant/torch_quant/quantizer.py
```diff
@@ -258,26 +234,48 @@ def qat_gm(
...
+        else:
+            ctx.act_ob_ctr = self.act_ob_ctr
+            ctx.w_ob_ctr = self.w_ob_ctr
+            ctx.bias_ob_ctr = self.bias_ob_ctr
+            ctx.modify_graph([set_qconfig, fuse_modules, insert_w_observer])
+
+            ctx.act_ob_ctr = ob_types.act_ob_ctr
+            ctx.w_ob_ctr = ob_types.w_ob_ctr
+            ctx.bias_ob_ctr = ob_types.bias_ob_ctr
+            ctx.modify_graph(
```
chenbohua3, line 255: why modify the graph twice?

tools/torch_quant/torch_quant/quantizer.py
```diff
@@ -258,26 +234,48 @@ def qat_gm(
...
+                [
+                    insert_act_observer,
+                    insert_w_observer,
```
chenbohua3, line 258: What is the purpose of insert weight observer alone instead of the origin one.

tools/torch_quant/torch_quant/quantizer.py
xiaowan0322: Reasonable. Done.

tools/torch_quant/torch_quant/quantizer.py
xiaowan0322: For reference and fbgemm backends, we didn't insert any `w_ob` or `bias_ob` in the `quantize` stage. Hence we should firstly insert normal `w_ob` or `bias_ob`, then replace to LSQObserver.

tools/torch_quant/torch_quant/quantizer.py
xiaowan0322: In early version, we only inserted weight observers for DISC backend, which was implemented in `quantizable_module_to_observed` pass. But now it is also needed by other backends for performing amp or qat.

chenbohua3 (approved):
LGTM

PR #1096, opened 2023-03-23
add flag to control multi-cc compilation

wyzero:
multi-cc compilation will take longer time. Add flag to control it.

Disable it will reduce the compilation time for `GpuKernelToBlobPass` from ~200s to ~80s.


qiuxiafei (approved):
LGTM
nit: the option could be shorter ?

qiuxiafei (approved):

PR #1143, opened 2023-05-16
Support acl bf16/fp16 gemm custom call

Guo-Peilin:


JamesTheZ:
Please add some UTs.

qiuxiafei:
> Please add some UTs.

but ...  passes are removed ... Orz ...

qiuxiafei (approved):
As discussed, this will be tested after #1176 .

PR #1164, opened 2023-05-26
Support to dump cuda ir and binary

bikekiller:
"export DISC_CUDA_KEEP_TEMPFILES=true" to dump
.ll, .ptx and .cubin for cuda backend.

### DIFF
--- a/tao_compiler/mlir/disc/transforms/disc_gpu_kernel_to_blob.cc
+++ b/tao_compiler/mlir/disc/transforms/disc_gpu_kernel_to_blob.cc
@@ -417,6 +417,51 @@ class GpuKernelToBlobPass
       TF_ASSIGN_OR_RETURN(gpu_asm,
                           tensorflow::se::BundleGpuAsm({image}, gpu_asm_opts));
     }
+
+    bool keep_tempfiles = false;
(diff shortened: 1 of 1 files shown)


tao_compiler/mlir/disc/transforms/disc_gpu_kernel_to_blob.cc
```diff
@@ -417,6 +417,53 @@ class GpuKernelToBlobPass
...
+    // Dump cubin.
+    std::ofstream fout(cubin_path, std::ios::binary);
+    fout.write(reinterpret_cast<const char*>(gpu_asm.data()), gpu_asm.size() * sizeof(uint8_t));
+    fout.close();
+
+    bool keep_tempfiles = false;
+    TF_CHECK_OK(tensorflow::ReadBoolFromEnvVar("DISC_CUDA_KEEP_TEMPFILES",
+                                                /*default_val=*/false,
+                                                &keep_tempfiles));
+    if (!keep_tempfiles) {
```
qiuxiafei, line 462: how about move `if` to L420 to guard all the code above? Because if one doesn't want to keep temp files, we even don't have to dump it.

bikekiller: Done

qiuxiafei (approved):
LGTM, Thanks!

PR #1187, opened 2023-06-08
Pass for transforming weight layout for A16w8/A16w4 Gemm

chenbohua3:


tao_compiler/mlir/disc/transforms/tests/disc-convert-weight-layout-for-weight-only-gemm.mlir
```diff
@@ -0,0 +1,39 @@
+// RUN: disc-opt --disc-transform-weight-data-layout-for-weight-only-quant -split-input-file %s | FileCheck %s
+
+
+func.func @convert_weight_layout_convert(%arg0: tensor<?x?x?xf16>) -> tensor<?x?x128xf16> {
+  // CHECK: weight_should_be_reordered = false
```
qiuxiafei, line 5: Do we need to check the layout is transformed correctly?

chenbohua3: It is not easy. The process of layout conversion requires the size of weight to be at least 64x2. The weight of this size will not show specific numbers on the mlir graph, which make it hard to check the transformed weight.

Maybe leave the check to the e2e test is better:)

qiuxiafei: sounds good.

qiuxiafei (approved):

PR #1221, opened 2023-07-05
support inplace kv cache compilation

Yancey1989:


qiuxiafei (approved):
LGTM... let's move on with fusion :P

PR #1227, opened 2023-07-17
remove stream handle for gpu memset

chenbohua3:


tao_compiler/mlir/ral/device/gpu/gpu_driver.h
```diff
@@ -77,7 +77,7 @@ class GPUDriver : public Driver {
   void d2h(ExecutionContext* ctx, stream_t, buffer_t from, buffer_t to, size_t);
   void h2d(ExecutionContext* ctx, stream_t, const_buffer_t from, buffer_t to,
            size_t);
-  void memset(ExecutionContext* ctx, stream_t, buffer_t, int, size_t);
```
qiuxiafei, line 80: Better to still retain the input stream? Referring [cuda manual](https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior__memset): 
> The cudaMemset functions are asynchronous with respect to the host except when the target memory is pinned host memory. The Async versions are always asynchronous with respect to the host.

`cudaMemset` a GPU memory seems to be async?  Do we need to synchronize it with other actions on the same CUDA stream?

qiuxiafei (approved):

PR #1282, opened 2024-02-22
Always try to lower standalone lmhlo.transpose to custom call for better performance

eedalong:
Always try to lower standalone lmhlo.transpose to custom call for better performance, no side effect.


Yancey1989 (approved):
LGTM

PR #1286, opened 2024-03-04
Support optimization barrier op

eedalong:
We expand `mhlo.optimization_barrier` after we've done graph level optimizations like fusion.

With this op, we can support training skills like GC and FSDP.


Yancey1989 (approved):
LGTM

PR #1304, opened 2024-06-19
Add DiscRematerializationPass to auto recompute

eedalong:


eedalong:
This pr is too large, close it and will have several small prs for op scheduling and recompute for dynamic shape & static shapes seperately.

